PR #15, opened 2017-08-12
Support route compilation (fixes #14)

### LINKED ISSUE #14: compile routes for snappy real-life use
When a `Match` is found, it should contain directly a optionally compiled handler to be invoked by the client.

ikitommi:
* also, so docs

### DIFF
--- a/src/reitit/core.cljc
+++ b/src/reitit/core.cljc
@@ -90,7 +90,7 @@
   "Creates a [[LinearRouter]] from routes and optional options.
   See [[router]] for available options"
   ([routes]
-    (linear-router routes {}))
+   (linear-router routes {}))
   ([routes {:keys [compile]}]
    (let [compiled (map (partial compile-route compile) routes)]
      (->LinearRouter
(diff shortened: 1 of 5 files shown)


src/reitit/core.cljc
```diff
@@ -55,18 +55,25 @@
...
 
-(defn resolve-routes [data {:keys [coerce] :or {coerce identity} :as opts}]
-  (->> (walk data opts)
-       (map-meta merge-meta)
-       (mapv (partial coerce))
-       (filterv identity)))
+(defn resolve-routes [data {:keys [coerce] :as opts}]
+  (cond-> (->> (walk data opts)
+               (map-meta merge-meta))
+          coerce (->> (mapv (partial coerce))
```
miikka: Could be simpler and maybe even more performant with `(cond->> ... coerce (into [] (keep (partial coerce))))`.

test/cljc/reitit/core_test.cljc
```diff
@@ -55,25 +55,44 @@
...
+                 (reitit/routes router))))
+        (testing "route match contains compiled handler"
+          (is (= 2 @compile-times))
+          (let [{:keys [handler]} (reitit/match-by-path router "/api/pong")]
+            (is handler)
+            (is (= "/api/pong" (handler)))
+            (is (= 2 @compile-times))))))
+    (testing "default compile"
+      (let [router (reitit/router ["/ping" (constantly "ok")])]
+        (println (reitit/match-by-path router "/ping"))
```
miikka: [nit] debug print

ikitommi: removed in ring-branch (on top of this)

src/reitit/core.cljc
ikitommi: 👍

PR #16, opened 2017-08-13
Ring router

ikitommi:
New `testit.ring` namespace, that has `router` and `ring-handler` functions to create a ring-router

## `simple-router`
* per route `:handler` is required (fails via custom `:compile` hook if missing)
* `:middleware` is stacked up from root
* no routing per request-method

```clj
["/api" {:middleware [api-mw]}
 ["/ping" handler]
 ["/pong" {:middleware [pong-mw]
           :handler handler}]]
```

## `method-router`
* like `simple-router` but routes per http-method
* top-level meta-data is meta-merged into method handlers
* custom `:coerce` and `:compile` hooks

```clj
["/api" {:middleware [api-mw]}
 ["/users" {:middleware [users-mw]
            :get {:middleware [get-mw]
                  :handler get-handler}
            :post post-handler
            :handler catch-others-handler}]]
```

### DIFF
--- a/src/reitit/core.cljc
+++ b/src/reitit/core.cljc
@@ -56,10 +56,8 @@
     {} x))
 
 (defn resolve-routes [data {:keys [coerce] :as opts}]
-  (cond-> (->> (walk data opts)
-               (map-meta merge-meta))
-          coerce (->> (mapv #(coerce % opts))
-                      (filterv identity))))
+  (cond->> (->> (walk data opts) (map-meta merge-meta))
+           coerce (into [] (keep #(coerce % opts)))))
 
 (defn compile-route [[p m :as route] {:keys [compile] :as opts}]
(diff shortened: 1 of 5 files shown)


miikka:
Does it make sense to have simple-router? You could just use method-router and never use the method-specific handlers, right?

ikitommi:
That's a valid point. And just for Ring (as all the naming suggest) this would be totally true. To support generic middleware (or interceptor) -based routers there should be the simple version, without any ties to the http. "Just mw & handlers" or "just interceptors & handlers". The latter could be usefull with re-frame for example.

Maybe there should be separate `reitit.interceptor` and `reitit.middleware` for the non-http case and `reitit.ring` would use the latter and have only the method-routing version.

What do you think?

miikka:
Right, okay. I don't really have an opinion on the naming.

PR #18, opened 2017-08-15
Match is injected into request

ikitommi:
Inject `Match` into request with `reitit.ring/router`. Also allow custom options to be passed to both `reitit.middleware/router` and `reitit.ring/router`. This enables for example easy way to mount shared :middleware for all routes. 

Together these enable that route meta-data can be interpreted also at runtime.

## Example

Middleware to enforce roles based on route meta-data (after route is matched):

```clj
(require '[reitit.ring :as ring])

(defn wrap-enforce-roles [handler]
  (fn [{:keys [::roles] :as request}]
    (let [required (some-> request (ring/get-match) :meta ::roles)]
      (if (and (seq required) (not (set/intersection required roles)))
        {:status 403, :body "forbidden"}
        (handler request)))))
```

Can be mounted for all routes:

```clj
(def handler (constantly {:status 200, :body "ok"}))

(def app
  (ring/ring-handler
    (ring/router
      [["/api"
        ["/ping" handler]
        ["/admin" {::roles #{:admin}}
         ["/ping" handler]]]]
      {:meta {:middleware [wrap-enforce-roles]}})))
```

Public route:

```clj
(app {:request-method :get, :uri "/api/ping"})
; {:status 200, :body "ok"}
```

Guarded route, without the roles:

```clj
(app {:request-method :get, :uri "/api/admin/ping"})
; {:status 403, :body "forbidden"}
```
Guarded route, with the needed roles:
```clj
(app {:request-method :get, :uri "/api/admin/ping", ::roles #{:admin}})
; {:status 200, :body "ok"}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ Path-based routing:
 
 On match, route meta-data is returned and can interpreted by the  application.
 
-Routers also support meta-data compilation enabling things like fast [Ring](https://github.com/ring-clojure/ring) or [Pedestal](http://pedestal.io/) -style handlers. Compilation results are found under `:handler` in the match. See [configuring routers](configuring-routers) for details.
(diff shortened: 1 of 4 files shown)


ikitommi:
true, compiling it's not.

PR #22, opened 2017-08-20
Small improvements

ikitommi:
* real-world perf-tests:
  * CQRS-sample from [Lupapiste](https://github.com/lupapiste)
  * REST-sample from [Opensensors](api.opensensors.io)
* perf tuning based on those:
  * `fast-map`, `fast-assoc` and `fast-get` for fast map access (on Clojure only for now)
* As lookup-tables are now backed by mutable Java HashMaps, `LinearRouter` and `LookupRouter` are replaced by reified Protocols to hide the impl details
* `router-type` & `options` to get public details out of Routers


perf-test/clj/reitit/opensensors_routing_test.clj
```diff
@@ -0,0 +1,563 @@
...
+
+(def opensensors-routes
+  [["/v2/whoami" {:handler handler, :name :test/route1}]
+   ["/v2/users/:user-id/datasets" {:handler handler, :name :test/route2}]
+   ["/v2/public/projects/:project-id/datasets" {:handler handler, :name :test/route3}]
+   ["/v1/public/topics/:topic" {:handler handler, :name :test/route4}]
+   ["/v1/users/:user-id/orgs/:org-id" {:handler handler, :name :test/route5}]
+   ["/v1/search/topics/:term" {:handler handler, :name :test/route6}]
+   ["/v1/users/:user-id/invitations" {:handler handler, :name :test/route7}]
+   #_["/v1/orgs/:org-id/devices/:batch/:type" {:handler handler, :name :test/route8}]
```
miikka: Why are some routes commented out?

src/reitit/core.cljc
```diff
@@ -60,69 +60,88 @@
   (cond->> (->> (walk data opts) (map-meta merge-meta))
            coerce (into [] (keep #(coerce % opts)))))
 
+(defn name-lookup [[_ {:keys [name]}] opts]
+  (if name #{name}))
+
+(defn find-names [routes opts]
+  (into [] (keep #(-> % second :name) routes)))
```
miikka: Could be marginally more efficient to use a transducer. `(into [] (keep #(-> % second :name)) routes)`

perf-test/clj/reitit/opensensors_routing_test.clj
ikitommi: There are conflicting routes and pedestals breaks with those. Also, pedestal fails on non-conflicting routes. Will write issue there.

src/reitit/core.cljc
ikitommi: :+1:

ikitommi:
Prefix-tree will always be slower than lookup-router so we optimize that too.

PR #24, opened 2017-08-21
Duplicate routes [WIP]

ikitommi:
Finds duplicate routes from a route tree. Should the `reitit.core/reitit` have an option to configure a callback to handle the duplicates? Maybe with "throw on duplicates" as default?


ikitommi:
throws by default, option to handle the side-effects

PR #25, opened 2017-08-21
Handle conflicting routes, custom & mixed-router

### LINKED ISSUE #23: Validate route-tree not to contain duplicate paths
Like the Pedestal prefix-tree router does. This would open up nice possibilities to optimise mixed trees with both wildcard & wildcard-free routes: the free-ones could be lookup-matched first. Also, if prefix-tree router is imported, we should test it's (ported) perf too. Might be best if tree has lot's of wildcard-routes...

ikitommi:
* Handle conflicting routes by `router`. by default, throws an exception, can be overridden with `:conflicts` option. Fixes #23
* new `:mixed-router` delegating to `:linear-router` or `lookup-router`, -20% faster on rest-sample
* ability to define custom routers via `router` options

Example exception message on conflicting routes:

```clj
CompilerException clojure.lang.ExceptionInfo: router contains conflicting routes:

   /v2/public/messages/dataset/bulk
-> /v2/public/messages/dataset/:dataset-id

   /v1/users/:user-id/devices/bulk
-> /v1/users/:user-id/devices/:client-id
```

### DIFF
--- a/src/reitit/core.cljc
+++ b/src/reitit/core.cljc
@@ -203,6 +203,37 @@
          (if-let [match (impl/fast-get lookup name)]
            (match params)))))))
 
+(defn mixed-router
+  "Creates two routers: [[lookup-router]] for static routes and
+  [[linear-router]] for wildcard routes. All routes should be
+  non-conflicting. Takes resolved routes and optional
+  expanded options. See [[router]] for options."
+  ([routes]
+   (mixed-router routes {}))
+  ([routes opts]
+   (let [{linear true, lookup false} (group-by impl/wild-route? routes)
+         linear-router (linear-router linear opts)
+         lookup-router (lookup-router lookup opts)
+         names (find-names routes opts)]
+     (reify Router
+       (router-type [_]
+         :mixed-router)
+       (routes [_]
+         routes)
+       (options [_]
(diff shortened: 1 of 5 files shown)


src/reitit/core.cljc
```diff
@@ -60,6 +61,27 @@
   (cond->> (->> (walk data opts) (map-meta merge-meta))
            coerce (into [] (keep #(coerce % opts)))))
 
+(defn conflicting-routes [routes]
```
miikka: This whole function might be more efficient and easier to understand with transducers. 😎

src/reitit/core.cljc
```diff
@@ -69,7 +91,7 @@
 (defn compile-route [[p m :as route] {:keys [compile] :as opts}]
   [p m (if compile (compile route opts))])
 
-(defprotocol Routing
+(defprotocol Router
```
miikka: This is a better name 👍

src/reitit/core.cljc
ikitommi: Before the Training I have no clue how to transduce this properly. Please help.

ikitommi:
added a comment about the transducer. ok to merge?

PR #39, opened 2017-10-02
Fast-router

ikitommi:
If only 1 static path in route tree, go with string match. 200x faster on basic benchmark than the slowest one. Why? To see how fast we can go.

`.equals` for String is 4x faster than `=` on the JVM. The `string_equals` has it's own HotSpot optimization, see https://github.com/openjdk-mirror/jdk7u-hotspot/blob/master/src/cpu/x86/vm/x86_64.ad#L10609


miikka:
`fast-router`  is maybe not the most descriptive name. Reitit is all about fast routers, after all! How about `single-path-router` or something like that?

ikitommi:
It only allows static paths, so renamed to `single-static-path-router`.

PR #40, opened 2017-10-26
module split + better docs

ikitommi:
Inspired by @mpenet, model from https://github.com/mpenet/alia

* `metosin/reitit` (all)
* `metosin/reitit-core` (just the router)
* `metosin/reitit-ring` (the ring stuff)
* `metosin/reitit-spec` (the spec coercion + spec-tools)

upcoming:

* `metosin/reitit-schema` (plumatic schema coericon)

maybe:

* `metosin/reitit-cljs`(cljs-routing helpers?)

to run lein commands for all modules:

```bash
./scripts/lein-modules install
```

* all tests are in the main project
* all sources are visible in the main project


ikitommi:
oh, the tests fail. I guess we would need to build the stuff locally before running those.

Deraen:
Why?

Reitit doesn't depend on anything (except meta-merge) and only benefit of modules I can think of is that modules can have their own dependencies.

ikitommi:
This is ahead of time PR.

I was about to add a Schema Coersion module, which will need the schema-dependency. Also, spec-tools is needed for the spec-coercion. Both could be separate coercion modules.

ikitommi:
Other ideas how to hadle this welcome!

project.clj
```diff
@@ -1,17 +1,20 @@
...
   :codox {:output-path "doc"
           :source-uri "https://github.com/metosin/reitit/{version}/{filepath}#L{line}"
           :metadata {:doc/format :markdown}}
 
-  :dependencies [[meta-merge "1.0.0"]]
+  :dependencies [[metosin/reitit-core ~reitit-version]
+                 [metosin/reitit-ring ~reitit-version]]
+
+  :source-paths ["modules/reitit-core/src"
+                 "modules/reitit-ring/src"]
```
miikka: Aren't source paths included in the JAR? So if you do release on the top-level, it will include sources for the modules.

lein-modules
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# A script that walks the tree of Leiningen projects and does $* on them
+# original: https://github.com/juxt/yada/blob/master/treelein
+
+# Modules
+for ext in reitit-core reitit-ring; do
+    cd modules/$ext; lein $*; cd ../..;
+    done
```
miikka: [nit] Weird indentation - `done` is usually indented to the same level as for.

lein-modules
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# A script that walks the tree of Leiningen projects and does $* on them
+# original: https://github.com/juxt/yada/blob/master/treelein
+
+# Modules
+for ext in reitit-core reitit-ring; do
+    cd modules/$ext; lein $*; cd ../..;
```
miikka: Should be `lein "$@"` instead of that `lein $*` so that in case anyone ever calls lein-modules with an argument that includes spaces, it works as it should.

lein-modules
```diff
@@ -0,0 +1,12 @@
...
+    done
+
+# Core
+lein $*
```
miikka: Ditto about `"$@"`

lein-modules
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# A script that walks the tree of Leiningen projects and does $* on them
+# original: https://github.com/juxt/yada/blob/master/treelein
+
```
miikka: Should you maybe `set -e` so that if any of the lein commands (or cds) fails, the script would stop?

modules/reitit-core/project.clj
```diff
@@ -0,0 +1,7 @@
+(load-file "../../.deps-versions.clj")
```
miikka: I hate it that Clojure project files have to be evaluated as code to be parsed. This is a language that promotos data-driven interfaces, after all!

modules/reitit-core/project.clj
```diff
@@ -0,0 +1,7 @@
...
+(defproject metosin/reitit-core reitit-version
+  :description "Snappy data-driven router for Clojure(Script)"
+  :url "https://github.com/metosin/reitit/tree/master/modules/reitit-core"
```
miikka: Maybe the URL should just point to reitit root? It will be shown in Clojars and elsewhere and I think that's where I'd like to go.

lein-modules
ikitommi: fixed

lein-modules
ikitommi: fixed

lein-modules
ikitommi: fixed

lein-modules
ikitommi: fixed

modules/reitit-core/project.clj
ikitommi: good point. Maven has declarative includes, maybe leiningen should too? 

```xml
  <parent>
    <groupId>org.codehaus.mojo</groupId>
    <artifactId>my-parent</artifactId>
    <version>2.0</version>
    <relativePath>../my-parent</relativePath>
  </parent>
```

and actually, it does have this: https://github.com/achin/lein-parent

ikitommi:
Ok, thanks for the comments. Setup was kinda bad. Now much better. With the use of [lein-parent](https://github.com/achin/lein-parent), all the child projects could be managed nicely without the impereative version settings - most likely.

## The question now remains: should we do this? 

I guess another option would be just to have guides "importing namespace `reitit.ring.coercion.schema` requires you to import artefacts : `[metosin/schema-tools "0.9.0"]` or above. Compojure-api does this right now. I'm not fond of that.

With the module split, having a deps to `metosin/reitit`, you get it all. `metosin/reitit-core` gives just 3 namespaces. Ring uses similar naming (and `lein-sub` instead of `./lein-modules`.

project.clj
ikitommi: moved under `:dev`, which fixed this.

ikitommi: and, I didn't think of this, THANKS.

modules/reitit-core/project.clj
ikitommi: fixed

ikitommi:
Well, this seems to work and most of the points are resolved. Version is still set via the `(def VERSION "0.1.0-SNAPSHOT")` in the `project.clj`s, updated via the `scripts/set-version` script. It could ahve with a better regex to update the versions in-place and the `def` could be removed.

Also, did some re-packaging, the ring-stuff is all under `reitit.ring` now.

I think this is the way to package the lib. 

Should we merge this?

ikitommi:
version is now updated in place. WIll merge this unless new comments pop up.

PR #49, opened 2017-11-20
Segment router (WIP)

ikitommi:
The prefix-tree router is broken.

The original (Pedestal)-version fails with the following paths:

```clj
[["/:kikka"]
 ["/api/ping"]
 ["/api/users/"]]
```

[By design](https://github.com/pedestal/pedestal/issues/532), the first route masks ALL OTHER ROUTES in the tree.

Reitit [prefix-tree](https://github.com/metosin/reitit/blob/master/modules/reitit-core/src/reitit/trie.cljc) fixes this but still fails on the following:

```clj
[["/:kikka"]
 ["/:kukka/ping"]
 ["/api/users/"]]
```

=> only one wildcard can be defined in a route node.

Thus, rethink and rewrite a wildcard-router for correctness. 

# segment-router

* splits paths into tokens between the `/` (both for the tree & request)
* reified protocols for managed mutability & speed

## performance

* 4x faster than Pedestal
* 30% faster than Reitit 

## tests

* much more tests than before.

screen shot 2017-11-24 at 19 00 31


ikitommi:
Added lot's of corner-case tests (for all wildcard-routers). One more to go and this will be merged. The `:prefix-tree-router` will be gone, it's not faster, is 4x lines and broken in many ways. Byebye.

ikitommi:
Well, I seem to merged the schema-coercion on top of this. While looking forward to the comments on the implementation, will merge this to master. It's all SNAPSHOT still.

PR #61, opened 2018-01-07
allow custom default-handler to ring-router

yogthos:
this change allows passing a custom default-handler such as a 404 handler

### DIFF
--- a/modules/reitit-ring/src/reitit/ring.cljc
+++ b/modules/reitit-ring/src/reitit/ring.cljc
@@ -18,8 +18,9 @@
 (defn ring-handler
   "Creates a ring-handler out of a ring-router.
   Supports both 1 (sync) and 3 (async) arities."
-  [router]
-  (let [default-handler (constantly nil)]
+  ([router]
+    (ring-handler router (constantly nil)))
+  ([router default-handler]
     (with-meta
       (fn
         ([request]


ikitommi:
Thanks for the PR! There was earlier discussion about the 404 and it got documented here: https://metosin.github.io/reitit/ring/ring.html#not-found. I now think passing in the not-found  handler might be a good as the function signature tells how to do this (opposed to "read the manual"). Especially with async it easy to get this wrong on the client side. 

There are at least three different conditions:

1) path doesn't match
2) method doesn't match
3) rejected - by code (e.g. mw or hander returns `nil`)

With quick look, I believe the PR only effects the 2. We have at least the options:

a) support all three cases separately (via options map?): more correct, more complex configuration
b) one handler for all cases: much simpler, good enough?
c) throw error with meta-data on exception - handler on single exception handler - frameworkish

What do you think? I'm leaning on b. Impl & tests for all 3 cases would be good to guard against potential regression.

ikitommi:
... and there is also

d) the not-found handler would be a `request match => response` function, so one could read out what was requested and what was matched so all cases 1-3  could be handler nicely in the callback. Default `(constantly {:status 404, :body "not found"})` would still work. but it could also be used to render a `{:status 405, :body "You tried PATCH-request to '/api/ping' to the method from ip... but the method is missing"}`

yogthos:
I think b) seems like a good approach as well. It would always be possible to wrap additional middleware to handle specific cases. The handler would get the original request, so if it needs to be clever it can still work with the request map to provide custom behavior.

ikitommi:
Let's do the b then. Also, as the match is injected into the request, I think all the three cases could be handled in the function. Do you have time to finalise this?

ikitommi:
Hi. I started need this too, and can implement this if you are busy.

yogthos:
Sounds like a plan. Sorry got pulled off, and kind of fell of my plate. :)

ikitommi:
np, did #63. Didn't know ho to effectively handle the `:not-acceptable` case with async ring. Hmm.

yogthos:
Using respond with it the way you're doing seems reasonable to me. What was your reservation with that?

PR #62, opened 2018-01-12
Frontend routing implementation (WIP)

Deraen:
Should the main package depend on these frontend modules?

### DIFF
--- a/examples/frontend/src/frontend/core.cljs
+++ b/examples/frontend/src/frontend/core.cljs
@@ -17,7 +17,10 @@
 (defn about-page []
   [:div
    [:h2 "About frontend"]
-   [:a {:href "http://google.com"} "external link"]])
+   [:ul
+    [:li [:a {:href "http://google.com"} "external link"]]
+    [:li [:a {:href (rfh/href @router ::foobar)} "Missing route"]]
+    [:li [:a {:href (rfh/href @router ::item)} "Missing route params"]]]])
 
 (defn item-page [match]
   (let [{:keys [path query]} (:parameters match)
--- a/examples/frontend/src/frontend/core.cljs
+++ b/examples/frontend/src/frontend/core.cljs
@@ -1,0 +1,50 @@
+(ns frontend.core
+    (:require [reagent.core :as r]
+              [reitit.core :as rc]
+              [reitit.frontend :as rf]
+              [reitit.frontend.history :as rfh]
+              [reitit.coercion.schema :as rs]))
+
+(def router (atom nil))
+
+(defn home-page []
+  [:div
+   [:h2 "Welcome to frontend"]
+   [:div [:a {:href (rfh/href @router ::about)} "go to about page"]]])
+
+(defn about-page []
+  [:div
+   [:h2 "About frontend"]
+   [:a {:href "http://google.com"} "external link"]
+   [:div [:a {:href (rfh/href @router ::frontpage)} "go to the home page"]]])
+
+(defonce match (r/atom nil))
+
+(defn current-page []
+  [:div
+   (if @match
+     [:div [(:view (:data @match))]])
+   (pr-str @match)])
+
+(def routes
+  (rc/router
+    [""
+     [""
+      {:name ::frontpage-root
+       :view home-page}]
+     ["/"
+      [""
+       {:name ::frontpage
+        :view home-page}]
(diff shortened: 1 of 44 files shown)


modules/reitit-frontend/src/reitit/frontend.cljs
```diff
@@ -0,0 +1,90 @@
...
+  [controller method]
+  (when-let [f (get controller method)]
+    (f (::params controller))))
+
+(defn- pad-same-length [a b]
+  (concat a (take (- (count b) (count a)) (repeat nil))))
+
+(defn apply-controllers
+  "Applies changes between current controllers and
+  those previously enabled. Resets controllers which
```
miikka: [nit] `which` -> `whose`

modules/reitit-frontend/src/reitit/frontend.cljs
```diff
@@ -0,0 +1,90 @@
...
+        match (or (reitit/match-by-path routes (.getPath uri))
+                  {:data {:name :not-found}})
+        q (query-params uri)
+        ;; Coerce if coercion enabled
+        c (if (:result match)
+            (coercion/coerce-request (:result match) {:query-params q
+                                                      :path-params (:params match)})
+            {:query q
+             :path (:param match)})
+        ;; Replace original params with coerced params
```
ikitommi: Rest of reitit keeps the coerced paramaters in a separate `:parameters` map, while retaining the originals. Would be coherent to do so here too.

modules/reitit-frontend/src/reitit/frontend.cljs
```diff
@@ -0,0 +1,90 @@
...
+                                      (pad-same-length new-controllers old-controllers))
+                                 (keep identity)
+                                 vec)]
+    (doseq [controller (map :old changed-controllers)]
+      (apply-controller controller :stop))
+    (doseq [controller (map :new changed-controllers)]
+      (apply-controller controller :start))
+    new-controllers))
+
+(defn hash-change [routes hash]
```
ikitommi: routes => router

modules/reitit-frontend/src/reitit/frontend.cljs
ikitommi: Should there be a side-effecting function `init!` or such with a callback function - to be called when the route changes. For Reagent, one could pass in `(partial swap! app-state assoc ::view)`. the re-frame would use this too.

modules/reitit-re-frame/src/reitit/re-frame.cljs
```diff
@@ -0,0 +1,111 @@
...
+      ::not-initialized)))
+
+(defn routed-view
+  "Renders the current view component.
+
+  Current view component is the :view from last :controllers entry with :view set."
+  []
+  (let [view @(re/subscribe [:routes/match])
+        ;; Select downmost controller that has view component
+        controller (first (filter :view (reverse (:controllers (:data view)))))]
```
ikitommi: Today I think the `:view` should not be part of the Controller, it should be normal route data instead. With the current solution, the last view masks all previous ones. If there are multiple controllers in an endpoint, the order matters. e.g.

```clj
["/admin/users" {:controllers [users-controller admin-controller]}]
```

```clj
["/admin/users" {:controllers [admin-controller users-controller]}]
```

might render totally different uis. Instead:

```clj
["/admin/users" {:view users-view
                 :controllers [users-controller admin-controller]}]
```

Deraen:
TODO: rename `reitit.frontend` to `reitit.browser`, update with latest code (e.g. query-string support) from app projects and separate controllers and other to other modules.

examples/frontend/src/frontend/core.cljs
```diff
@@ -0,0 +1,50 @@
...
+
+(defonce match (r/atom nil))
+
+(defn current-page []
+  [:div
+   (if @match
+     [:div [(:view (:data @match))]])
+   (pr-str @match)])
+
+(def routes
```
ikitommi: looks complex. would something like this work:

```clj
(def routes
  (reitit.core/router
    [["/"
      {:name ::frontpage
       :view home-page}]
     ["/about"
      {:name ::about
       :view about-page}]]))
```

Deraen: No.

PR #63, opened 2018-01-23
Support not-found with ring

ikitommi:
Implemented the pluggable `default-handler` for ring, with a generator for  `:not-found`,  `:method-not-accepted` & `:not-acceptable`

Completes #61.


ikitommi:
Run some of the perf tests too, doesn't show on the real life tests.

PR #70, opened 2018-02-20
Added "Dev Workflow" to "Advanced" docs

### LINKED ISSUE #69: Support Vars in router
I'm trying to have my router dynamically update when I recompile the sub-routes in another namespace.
I thought that using a var might give me that but it looks like it doesn't.

```clojure
(r/match-by-path (r/router 
                    ["/foo"
                     #'m/router])
                   path)
```
No implementation of method: :expand of protocol:
   #'reitit.core/Expand found for class: clojure.lang.Var

I'm not sure this makes sense at all. But that's a concern that might maybe be adressed.

cark:
Closes #69

All example code was tested.
I'm not a native English speaker, please review and correct before merging.

Signed-off-by: Cark <carkhy@gmail.com>

### DIFF
--- a/doc/SUMMARY.md
+++ b/doc/SUMMARY.md
@@ -18,6 +18,7 @@
     * [Configuring Routers](advanced/configuring_routers.md)
     * [Different Routers](advanced/different_routers.md)
     * [Route Validation](advanced/route_validation.md)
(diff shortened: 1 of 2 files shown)


ikitommi:
back, thanks for this!

ikitommi:
i'll just merge this, might polish it later.

PR #73, opened 2018-03-13
Migrate all tests from Travis to Circle

miikka:


miikka:
Travis fails because I remove .travis.yml :P

PR #76, opened 2018-03-20
Initial take on IntoString

ikitommi:
Transforms given path-parameters via `reitit.impl/IntoString` protcol.

```clj
(r/match-by-name
  (r/router
    ["/olipa/:kerran" ::avaruus])
  ::avaruus
  {:kerran :joskus})
;#Match{:template "/olipa/:kerran",
;       :data {:name :user/avaruus},
;       :result nil,
;       :path-params {:kerran "joskus"},
;       :path "/olipa/joskus"}
```

### DIFF
--- a/modules/reitit-core/src/reitit/impl.cljc
+++ b/modules/reitit-core/src/reitit/impl.cljc
@@ -169,8 +169,8 @@
           (.replace "+" "%20")))
 
 (defn url-decode [s]
-  #?(:clj  (some-> s (URLDecoder/decode "UTF-8"))
-     :cljs (some-> s (js/decodeURIComponent))))
+  (some-> s #?(:clj  (URLDecoder/decode "UTF-8")
+               :cljs (js/decodeURIComponent))))
 
 (defprotocol IntoString
   (into-string [_]))
@@ -183,9 +183,8 @@
   #?(:clj  clojure.lang.Keyword
      :cljs cljs.core.Keyword)
   (into-string [this]
-    (str (namespace this)
-         (when (namespace this) "/")
-         (name this)))
+    (let [ns (namespace this)]
+      (str ns (if ns "/") (name this))))
 
   #?(:clj  Boolean
      :cljs boolean)
--- a/modules/reitit-core/src/reitit/impl.cljc
+++ b/modules/reitit-core/src/reitit/impl.cljc
@@ -166,7 +166,7 @@
   (some-> s
           #?(:clj  (URLEncoder/encode "UTF-8")
              :cljs (js/encodeURIComponent))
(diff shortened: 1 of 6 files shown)


modules/reitit-core/src/reitit/impl.cljc
```diff
@@ -155,3 +157,53 @@
 
 (defn strip-nils [m]
   (->> m (remove (comp nil? second)) (into {})))
+
+;;
+;; Path-parameters, see https://github.com/metosin/reitit/issues/75
```
miikka: The issue still doesn't explain what this is about, there's only discussion about a possible implementation. It'd be better to have a short explanation here.

modules/reitit-core/src/reitit/impl.cljc
```diff
@@ -155,3 +157,53 @@
...
+;;
+
+(defn url-encode [s]
+  (some-> s
+          #?(:clj  (URLEncoder/encode "UTF-8")
+             :cljs (js/encodeURIComponent))
+          (.replace "+" "%20")))
```
miikka: This does the wrong thing in JavaScript:

```
cljs.user=> (.replace "aaa" "a" "b")
"baa"
```

OTOH it's only needed for Clojure, since `js/encodeURIComponent` encodes space ` ` with `%20` anyway. For clarity, I'd call `.replace` only for Clojure.

modules/reitit-core/src/reitit/impl.cljc
```diff
@@ -155,3 +157,53 @@
...
+  (into-string [this] (str this))
+
+  #?(:clj  Object
+     :cljs object)
+  (into-string [this] (str this)))
+
+(defn path-params
+  "shallow transform of the path-param values into strings"
+  [params]
+  (persistent!
```
miikka: Did you benchmark using transients here? Because I recall that transients for small maps (TransientArrayMap/PersistentArrayMap) are slower than persistents, because Rich Hickey has optimized the latter but not applied the same optimizations to the former.

(Which should be fixed, but I've never gotten around to opening an issue or submitting a patch.)

ikitommi: Oh, true that. non-transient is faster with small sizes, which here is the case

modules/reitit-core/src/reitit/impl.cljc
ikitommi: awesome catch!

modules/reitit-core/src/reitit/impl.cljc
ikitommi: added

Issue #77, opened 2018-04-20
URL Query and Fragment Strings break match-by-path


SVMBrown:
Query and fragment strings cause wierd values for path parameters at the end of URLs, and cause routes to fail to match. In my opinion, path and fragment strings should either be coerced as `:query-params` and `:fragment-id`, extracted as `:query-string` and `:fragment-string`, or at least trimmed and ignored prior to matching. `:path-params` shouldn't include the query string appended to the final parameter. I am unsure of whether the `:path` key should trim or retain the query string.

``` clojure
(reitit/match-by-path 
  (reitit/router [["/foo" :foo] ["/bar/:id" :bar]]) 
  "/bar/100?a=b&c=d#foo")
;; Returns:
#_
#reitit.core.Match{:template "/bar/:id",
                   :data {:name :bar},
                   :result nil,
                   :path-params {:id "100?a=b&c=d#foo"},
                   :path "/bar/100?a=b&c=d#foo"}

```

``` clojure
(reitit/match-by-path 
  (reitit/router [["/foo" :foo] ["/bar/:id/baz" :bar]])
  "/bar/100/baz?a=b&c=d#foo")
;;Returns:
#_
nil
```

Deraen:
Frontend module now reads query strings and ignores the fragment: https://github.com/metosin/reitit/blob/master/test/cljs/reitit/frontend/core_test.cljs#L70-L90

On server-side Ring obviously handles query string and server shouldn't ever see fragment as it is not part of the request url.

I think we can close this issue now? Or should the fragment extracted into the match?

PR #78, opened 2018-04-20
Handle queries and fragments in URL

SVMBrown:
Fixing #77 
To solve handling of queries and fragments in a provided URL, I added a pre-process step that splits the URL in three parts and returns a map containing the trimmed path and the fragment and query strings if present. I am unsure how query params should fit into the larger API of parameter coercion, so I have left the query unparsed.

I understand that the duplication of merge logic with matches across each implementation of the Router protocol is less than ideal, but wanted to make as small a change to the existing API as possible. If you have an idea of how to better support URL trimming I am happy to help out with implementing it.

### DIFF
--- a/modules/reitit-core/src/reitit/core.cljc
+++ b/modules/reitit-core/src/reitit/core.cljc
@@ -164,12 +164,17 @@
          opts)
        (route-names [_]
          names)
-       (match-by-path [_ path]
-         (reduce
-           (fn [_ ^Route route]
-             (if-let [path-params ((:matcher route) path)]
-               (reduced (->Match (:path route) (:data route) (:result route) path-params path))))
-           nil pl))
+       (match-by-path [_ raw-path]
+         (let [{:keys [path query-string fragment-string] :as path-data}
+               (impl/pre-process-path raw-path)]
+           (when-some [match (reduce
+                              (fn [_ ^Route route]
+                                (if-let [path-params ((:matcher route) path)]
+                                  (reduced (->Match (:path route) (:data route) (:result route) path-params path))))
+                              nil pl)]
+             (merge match
+                    path-data
+                    {:path raw-path}))))
        (match-by-name [_ name]
          (if-let [match (impl/fast-get lookup name)]
            (match nil)))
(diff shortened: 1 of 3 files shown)


ikitommi:
Thanks for the PR!!

Do you need this on JVM, node, browser or on something else? 

Some thoughts on targets:

### JVM/Server

Http-adapters already parse the query-string out of an url and push both `:uri` and `:query-string` into the request. So, there is no need to always re-parse the uri to look for fragments. Reitit wants to stay as fast as possible, so no extra work should be done. There are existing (ring) middleware & interceptors to push the `:query-params` into the request.

### Browser

Javascript & Closure have optimized(?) tools for parsing the fragments. These could be applied before actual routing is called and injected into `Match`, just like Coercion.  We are doing this in our projects and @Deraen started to push some tools a [separate module](https://github.com/metosin/reitit/pull/62/files), unfinished thou.

We could take the (unfinished) keechma-style controllers and re-frame stuff away from the module there and just push out the helpers to plug reitit easily. Would that solve this issue? 

### Node

I don't know what is a best way to handle these in node.

cheers,

Tommi

SVMBrown:
Hi Tommi, I am currently using reitit as a router for kee-frame (a keechma-like layer on top of re-frame). I could add the query and fragment parsing as a wrapper to my project for sure. If that is what you recommend though, a note about query strings on the readme would be appreciated, as I was quite surprised to find that `/foo/123?a=b` matched on `/foo/:id` gives me `"123?a=b"` as `:id` when most other routing libs at least ignore the query string for similar cases. Having those modules when they're finished will help as well. Thanks for the library! Enjoying working with it so far!

tkachenko1503:
Hello guys! Just faced with troubles that will be solved by this PR. When do you plan to release it?

ikitommi:
Hi. @Deraen will finalize the #62 which will solve this effectively. I'll close this when it's done so everyone get's noticed. You can copy-paste the current parsing codes from there if you need it now, hopefully released within a week.

tkachenko1503:
thanks. I will wait for a release

ikitommi:
#62 has been merged to master, adding the helpers to do this. Closing this now.

PR #79, opened 2018-04-22
Ring perf and serve static resources

ikitommi:
* coherent style for ring-helpers `swagger/swagger-spec-handler` => `swagger/create-swagger-handler`
* fix schma + swagger tests
* 192n => 132ns for `ring-app` perf via code polishing
* ring/routes to help build default-ring handlers for both 1- & 3-arity. Copied from compojure
* ring/mime helpers, copied from ring, but much faster: 629ns => 106ns 
* `create-resource-handler` to serve static files (just clj)


CHANGELOG.md
```diff
@@ -19,9 +19,15 @@
 ;       :path "/coffee/luwak"}
 ```
 
+### `reitit-ring`
+
+* `reitit.ring/default-handler` now works correctly with async ring
+* new helper `reitit.ring/router` to compose routes outside of a router.
+* `reitit.ring/create-resource-handler` function to serve static routes. See [https://metosin.github.io/reitit/ring/static.html](docs).
```
miikka: [minor] Markdown link syntax is the wrong way around

doc/ring/static.md
```diff
@@ -0,0 +1,89 @@
...
+(require '[reitit.ring :as ring])
+
+(ring/ring-handler
+  (ring/router
+    [["/ping" (constantly {:status 200, :body "pong"})]
+     ["/assets/*" (ring/create-resource-handler)]])
+  (ring/create-default-handler))
+```
+
+To serve static files with conflicting routes, e.g. `"/*#`, one needs to disable the confligt resolution:
```
miikka: Typo: `confligt` -> `conflict`

doc/ring/static.md
```diff
@@ -0,0 +1,89 @@
...
+| key          | description |
+| -------------|-------------|
+| :parameter   | optional name of the wildcard parameter, defaults to unnamed keyword `:`
+| :root        | optional resource root, defaults to `"public"`
+| :mime-types  | optional extension->mime-type mapping, defaults to `reitit.ring.mime/default-types`
+| :path        | optional path to mount the handler to. Works only if mounted outside of a router.
+
+### TODO
+
+* support for things like `:cache`, `:last-modified?` and `:index-files`
```
miikka: Another cool thing could be serving gzipped versions of the files.

doc/ring/static.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Performance
```
miikka: It's a bit unclear to me if this perf section is useful. There's no point of comparison for the numbers below and anyway a single number isn't very informative – usually you'd want to look at the distribution.

Besides, I tried running the large file command against my web app that uses Aleph, `ring.util.response/resource-response` and custom regexp-based routing on my mid-2015 MBP and got this:

```
Requests/sec:   6398.48
Transfer/sec:      2.48GB
```

I didn't try reitit on my computer, but just based on the text here I would've expected this benchmark to show higher numbers than what I'd get with my setup.

If you'd like to include a point of comparison, I think something like nginx might be a good one. It's the de facto server for serving static files, after all.

modules/reitit-ring/src/reitit/ring.cljc
```diff
@@ -48,6 +67,46 @@
...
+                        :headers {"Content-Type" (mime/ext-mime-type (.getName file) mime-types)}})]
+        (if path
+          (let [path-size (count path)]
+            (fn
+              ([req]
+               (let [uri (:uri req)]
+                 (if (and (>= (count uri) path-size))
+                   (some->> (str root (subs uri path-size)) io/resource io/file response))))
+              ([req respond _]
+               (let [uri (:uri req)]
```
miikka: Could simplify by naming this function with `(fn foo ...)` and then here `(respond (foo req))`.

ikitommi:
Polished the impl: uses now ring `resource-response` to serve resources: works also from uberjar and from there, it's only marginally slower as files are sent using non-nio streams anyway. Now the "Last-modified" is also set.

PR #87, opened 2018-05-20
fix various links in documentation

martinklepsch:
These probably work in Gitbook to some extent but they don't work on
GitHub. In most places of the documentation files are referenced
via their source files (`.md`) and I assume Gitbook deals fine with that

For cljdoc the correct links are required to properly fix links that are
broken when rendering the document on different URLs.


ikitommi:
Gitbook deals with those fine, just typos. Thanks!

PR #91, opened 2018-05-28
Fix paths in swagger.json for endpoints with path parameters

DeLaGuardo:


modules/reitit-swagger/src/reitit/swagger.cljc
```diff
@@ -62,6 +63,21 @@
   {:name ::swagger
    :spec ::spec})
 
+(defn- path->template [path endpoint]
```
ikitommi: This could be simplified to:

```clj
(require '[reitit.impl :as impl])

(defn- path->template [path]
  (->> (impl/segments path)
       (map #(if (impl/wild-or-catch-all-param? %)
               (str "{" (subs % 1) "}") %))
       (string/join "/")))
```
, making it support also routes that don't have the `:path` parameters defined.

ikitommi: Actually, I believe Swagger doesn't support catch-all-params, so a route `["/files/*" ...]` would not work. Maybe we should strip those away:

```clj
(defn path->template [path]
  (let [segments (impl/segments path)]
    (if-not (some impl/catch-all-param segments)
      (->> segments
           (map (fn [param]
                  (if-let [wild-param (impl/wild-param param)]
                    (str "{" wild-param "}")
                    param)))
           (string/join "/")))))
```

what do you think?

DeLaGuardo: Yes, catch-all-params is not supported by swagger. So impl/wild-param should be enough.

DeLaGuardo: As far as I understand swagger doesn't care about content of path parameter and it can be anything. `some/part/of/the/path` as example. That is exactly the catch-all-params idea. So I would rather choose your first variant cause it looks nicer)

ikitommi:
Awesome, thanks!

PR #96, opened 2018-06-05
0.1.2

ikitommi:
... with these extra fixes, in this PR:

```markdown
### `reitit-core`

* Better handling of `nil` in route syntax:
  * explicit `nil` after path string is always handled as `nil` route
  * `nil` as path string causes the whole route to be `nil`
  * `nil` as child route is stripped away

(testing "nil routes are stripped"
  (is (= [] (r/routes (r/router nil))))
  (is (= [] (r/routes (r/router [nil ["/ping"]]))))
  (is (= [] (r/routes (r/router [nil [nil] [[nil nil nil]]]))))
  (is (= [] (r/routes (r/router ["/ping" [nil "/pong"]])))))

### `reitit-ring`

* Use HTTP redirect (302) with index-files in `reitit.ring/create-resource-handler`.
* `reitit.ring/create-default-handler` now conforms to [RING Spec](https://github.com/ring-clojure/ring/blob/master/SPEC), Fixes [#83](https://github.com/metosin/reitit/issues/83)
```

### DIFF
--- a/modules/reitit-core/src/reitit/core.cljc
+++ b/modules/reitit-core/src/reitit/core.cljc
@@ -37,16 +37,17 @@
     [(walk-many [p m r]
        (reduce #(into %1 (walk-one p m %2)) [] r))
      (walk-one [pacc macc routes]
-       (if (vector? (first routes))
-         (walk-many pacc macc routes)
-         (let [[path & [maybe-arg :as args]] routes
-               [data childs] (if (vector? maybe-arg)
-                               [{} args]
-                               [maybe-arg (rest args)])
-               macc (into macc (expand data opts))]
-           (if (seq childs)
-             (walk-many (str pacc path) macc childs)
-             [[(str pacc path) macc]]))))]
+       (if-let [routes (seq (keep identity routes))]
+         (if (vector? (first routes))
+           (walk-many pacc macc routes)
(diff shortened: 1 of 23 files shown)


modules/reitit-core/src/reitit/core.cljc
```diff
@@ -39,14 +39,14 @@
...
          (walk-many pacc macc routes)
-         (let [[path & [maybe-arg :as args]] routes
-               [data childs] (if (vector? maybe-arg)
-                               [{} args]
-                               [maybe-arg (rest args)])
-               macc (into macc (expand data opts))]
-           (if (seq childs)
-             (walk-many (str pacc path) macc childs)
-             [[(str pacc path) macc]]))))]
+         (if (string? (first routes))
```
kalekale: Maybe when instead? I at least instinctively look for the third argument in ifs.

ikitommi: There is no consensus on this: `when` has implicit `do` for side-effects, so one needs to look for the last statement in the block to see what it returns and see if it side-effects. `if` always just returns.

But, changed this to `when`.

related: https://github.com/jonase/kibit/issues/2

ikitommi:
Thanks for the review!

PR #110, opened 2018-07-17
Frontend small changes

ikitommi:
* `coercion/coerce!` can be used now
* cleanup imports


ikitommi:
As a side-note, the `reitit.impl/goog-extend` macro should be in the `reitit-frontend` module, as it only works there.

Issue #84, opened 2018-05-14
Support for OpenAPI3


ikitommi:
This spans over multiple repos, but can be done in parts.

## Schema -> OpenAPI3
* https://github.com/metosin/schema-tools/issues/44
* might be some work to get this clean without copying a lot

## Spec -> OpenAPI3
* https://github.com/metosin/spec-tools/issues/105
* should be easy, a new ns just like `spec-tools.swagger` overriding some JSON Schema parts (if any!)

## Malli -> OpenAPI3
* https://github.com/metosin/reitit/blob/master/modules/reitit-malli/src/reitit/coercion/malli.cljc

## Reitit Coercion
* currently, there is only one body-coercion for request and one for response. We should add support for content-type based coercion:

### Coercion for all content-types

```clj
(require '[reitit.ring :as ring])
(require '[reitit.ring.coercion])
(require '[reitit.coercion.spec])
(require '[muuntaja.middleware])

(ring/ring-handler
  (ring/router
    ["/plus"
     {:get {:parameters {:body {:x int?, :y int?}}
            :responses {200 {:body {:total pos-int?}}}
            :handler (fn [{{{:keys [x y]} :body} :parameters}]
                       {:status 200, :body (+ x y)})}}]
    {:data {:middleware [muuntaja.middleware/wrap-format
                         reitit.ring.coercion/coerce-request-middleware
                         reitit.ring.coercion/coerce-response-middleware]
            :coercion reitit.coercion.spec/coercion}}))
```

### Coercion per content-type

* new type `:content` (from OpenAPI3) which allows content-type->model mappings.
* should be disallowed with Swagger

```clj
(ring/ring-handler
  (ring/router
    ["/plus"
     {:get {:parameters {:content {"application/json" {:x neg-int?, :y neg-int?}
                                   "application/edn" {:x pos-int?, :y pos-int?}}}
            :responses {200 {:content {"application/json" {:json int?}
                                       "application/edn" {:edn int?}}}}
            :handler (fn [{{{:keys [x y]} :body} :parameters :as request}]
                       (let [content-type (-> request :muuntaja/response :format)
                             body (case content-type
                                    "application/json" {:json (+ x y)}
                                    "application/edn" {:edn (+ x y)})]
                         {:status 200, :body body}))}}]
    {:data {:middleware [muuntaja.middleware/wrap-format
                         reitit.ring.coercion/coerce-request-middleware
                         reitit.ring.coercion/coerce-response-middleware]
[...]

PR #114, opened 2018-07-22
Reitit default middleware

ikitommi:
new module `metosin/reitit-middleware` with common / handy ring-middleware, aiming to be on par with compojure-api.

Review the decisions:

# exception-middleware

* in compojure-api, the middleware took options map with just `:handlers` key. As there are no other top-level keys, the handlers-map is the only arg taken. I think this is better this way. There is also new `:reitit.ring.middleware.exception/wrap` key  which can be used for example to log all errors - like ring middleware, but for exception handlers, takes 3 args instead of the normal 2 (handler being the first one). With the old options-syntax, this could be the other top-level key in options besides the `:handlers`, but a special key gives better ergonomics I think.

# multipart-middleware

* ok'ish, before the arrival of https://github.com/metosin/reitit/issues/84

# muuntaja

* the middleware do not take any options, but they read the `:muuntaja` key from route data. If that is not set, nothing happens (without any warning!)
   * => one needs to set both the middleware and the muuntaja instance to get it working. We could set the muuntaja instance to default one if it's not set, but that would mean it's always enabled for all paths. Currently [meta-merge doesn't support removing keys easily](https://github.com/weavejester/meta-merge/issues/11) -  if we inlined meta-merge, we could use the default muuntaja instance if that is not explicitly set  (the common case) and people could set it to nil for certain routes which don't require the functionality.
   => I think the explicit `:muuntaja` key is good.

# naming

* module: `reitit-middleware` or `reitit-ring-middleware` or something different?
* ns'es: `reitit.ring.middleware.exception` or just `reitit.middleware.exception`?


ikitommi:
reviewed by @jarppe

PR #115, opened 2018-07-23
Add URL-decoding support for paths and params

KingMob:
Hi, we chatted briefly about this in Slack a couple weeks ago, and you said you'd welcome a PR, so here it is. I tried to make minimal changes to get it working, but a lot of the code for constructing a route is the same as for matching a URL, even if they should be treated differently when encoding is present. The tests all pass, though.

Summary:
Currently Reitit cannot handle %-encoded paths or params, even though they can be
quite common (like `/search/random%20search%20string`) This PR enables matching URLs with %-encoded chars, and decodes path params.

After applying:
- Successfully match `/space in path` to `/space%20in%20path`
- Match `/path/:param1` to `/path/foo%20bar` and see
`:param1` => `"foo bar"` in path params

Caveats:

- Some of the routing does whole-URL regex matching which is not quite correct, but errors are probably rare. Fixing this would require decoding each path part separately, though, and would be a bit slower. (E.g., after decoding, `/path/to%2Fthing` can match `/path/to/thing` with certain router types, which is not actually correct.)
- Is inconsistent when decoding '+' into a space for path parts, because the Java and Js fns disagree. (Java's URLDecoder supports it, but it's technically for x-www-form-urlencoded data, and not URLs. Js's decodeURIComponent doesn't support it, which is correct for URL path parts, though not URL query parts.) If you want to pick a side to fall on, I'm ok with updating it to support that.


ikitommi:
Hi. And thanks for the PR! 

I had to read how the url-encoding should actually work. Agree that the original use of `URLEncoder` and `URLDecoder` is not correct. Versions in `ring.util.codec` are bit better, but still work differently than with JavaScript and would add an extra dependency. There is a decent version in both Play Framework and in Spring Framework - and a differently working version in Guava.

While my quick study, I made a almost-working version of encode & decode with tests from Play, up to 10x faster than the ring-version and seems to give more correct results. I'll push those as a separate commit. I guess we need separate `url-encode|decode` (path parameters) and `form-encode|decode` (query-params)...

About the PR.

Performance - I think the segment-router will be an order of magnitude slower if all the path segments are decoded on `-lookup`. Better would be just to collect a list of (wild) parameter key-value tuples and in case of an exact match, the params would be decoded. This way, no decoding is done for all tested paths (invalid paths are walked some steps before a dead-end is found). I think similar approach would be good for linear-router too.

Static path segments - hopefully they could either be not touched at all (like Compojure does, e.g. `"/olipa kerran/:id"` will never match to anything coming from a ring adapter, which sends all paths as encoded, `"/olipa%20kerran/123"`) or they could be pre-processed (recursively/fully decoded  + encoded once?) on router creation so that there would be no encoding/decoding for them on lookup time. With reverse routing, we need to ensure all segments are encoded, but that's not a perf critical track.

What do you think?

ikitommi:
PR for the better `encode` & `decode` functions: https://github.com/metosin/reitit/pull/120

ikitommi:
encode & decode are now (more) perf optimized for the critical path. And tested to work the same way with both JVM & JS.

KingMob:
Cool. I won't get a chance to really look at this again until Sunday at the earliest, but off-hand, I definitely agree it would be better to only decode parameter segments after matching. 

Non-parameter segments that require decoding are probably rare (e.g., who manually makes routes like `/path/to/my route with spaces/` anyway?), but I agree that pre-processing is the way to go. After all, if we consider that these are matched against real URLs, the spec is clear that you can't have a space in the path, so strictly speaking, something like `/olipa kerran/` is unmatchable.

I think I'll be able to take a crack on Sunday, but if not then, sometime next week, if you're in no hurry.

ikitommi:
Got bitten by this too while porting a large app to reitit, some `ö` and `ä` in the path parameters. Do you think you have time this week to check this? I think first version could be just to decode path-parameters as last step of `match-by-path` and inside ´match-by-name`. I can help if you don't have time.

KingMob:
I got something working on Sunday for params, but then I got hung up how/whether we should convert un-encoded paths supplied for initialization; e.g., if users supply "/my/spaced route", we should ideally convert it to  "/my/spaced%20route" for them, but if they pre-encode it, we shouldn't then turn "/my/spaced%20route" into "/my/spaced%2520route". 

We could try auto-detecting percent-encodings, though that will be error-prone (in a URL like "/products/vitamins/fish-oil-max-100%benefit", "%be" is valid hex. And surely a user will produce a mix of both encoded and unencoded one day.

For now, if you like, I can submit just the code for decoding the params, and we can deal with encoding supplied URLs separately (Slack, email, new PR, I'm flexible)?

ikitommi:
Decoding params PR most welcome :)

KingMob:
OK! Since this PR's title originally just referred to decoding, I pushed all the path-param-decoding code here, and will open a new PR for encoding supplied routes. Hope that's ok, but if not, lemme know.

I added a `map-kv` utility fn, refactored `path-params` to use it, and extracted the lookup structure-generating code (the ones that create `pl` and `nl` in the router fns) into their own fns because I was having a devil of a time figuring that code out. I hope they clarity, but if you hate them, they're in their own commit, so we can ditch it.

The key changes are in ec051a0. I reset the code a bit and then changed it so that for the linear and segment routers only, `match-by-path` will decode the final `path-params`. Since no other router can have params, that should work.

I updated the tests, and they all pass for both clj and cljs.

ikitommi:
Beautiful. Thanks!!

PR #120, opened 2018-08-01
Encode decode

ikitommi:
* `reitit.impl/url-encode`
* `reitit.impl/url-decode`
* `reitit.impl/form-encode`
* `reitit.impl/form-decode`

All work identically for JVM & JS. Much faster impl, more tests. Common case of non-esaped path-parameter is decoded 10x faster the `ring.util.codec` versions. 

Query-params are now encoded with `form-encode`, making spaces into `+` instead of `%20`.


modules/reitit-core/src/reitit/impl.cljc
```diff
@@ -163,18 +163,59 @@
...
+      4 "4" 5 "5" 6 "6" 7 "7"
+      8 "8" 9 "9" 10 "A" 11 "B"
+      12 "C" 13 "D" 14 "E" 15 "F"}))
+
+#?(:clj
+   (defn byte->percent [byte]
+     (let [byte (bit-and 0xFF byte)
+           low-nibble (bit-and 0xF byte)
+           high-nibble (bit-shift-right byte 4)]
+       (str "%" (hex-digit high-nibble) (hex-digit low-nibble)))))
```
miikka: Meh, Github ate my comment here. Here it is again:

This looks like a lookup table... Why not go further and have a bigger lookup table?

```clj
(def +percents+ (into-array (map #(format "%%%02x" %) (range 255))))
(defn byte->percent [byte] (aget +percents+ byte))
```

ikitommi: brilliant.

ikitommi: except:

```
expected: (= "%E2%80%9C%D0%8C%CF%8D%D0%91%D0%87%E2%80%9D" (impl/url-encode "“ЌύБЇ”"))
  actual: java.lang.ArrayIndexOutOfBoundsException: -30
```

miikka: Right, I forgot about Java's signed bytes.

```clj
(defn byte->percent [byte] (aget +percents+ (bit-and 0xFF byte)))
```

ikitommi:
```clj
expected: (= "%2F" (impl/url-encode "/"))
  actual: (not (= "%2F" "%ffffffffffffff30"))
```

ikitommi:
fixed. it's double fast (400ns -> 200ns), but the regexp is still dominating the conversion. It's not on perf critical path, can be made better later.

ikitommi:
String-array vs PersistentVector didn't show up here, but String array type hint is bit gory.

PR #121, opened 2018-08-08
URL-encode supplied routes?

KingMob:
## URL-encoding supplied routes

At the moment, this should probably be more of a discussion. The initially committed code will happily convert your routes, but runs into issues if *you've* already converted those routes, since it'll double-encode the %'s (turning `/spaced%20path` into `/spaced%2520path`). What to do in this situations is not exactly clear. I see several options:

1. Always encode the routes. Users will have to be educated to never encode the routes themselves. Probably fine for hand-written routes, but I can easily imagine tool/DB-generated routes already being encoded.
2. Never encode the routes. Users will have to know about this, too, but a lot of the time, cursory debugging will reveal the issue if they run into errors. Has the appeal of *forcing* users to supply valid routes, but less friendly. I suspect most users shy away from reserved chars to be safe, so this is probably ok.
3. Add a flag. Set a default, but allow users to change it, either globally or per route. Adds complexity but flexibility.
4. Try to determine if percent-encoding is present on behalf of the user. User-friendly, but impossible to always get right without, say, natural-language processing, and probably not even then. (E.g., how do you handle `/100%best-product`?)

I'm leaning towards 2 or 3. 2 is zero work, other than maybe updating the docs, or issuing a warning if reserved chars are encountered.

Thoughts?

### DIFF
--- a/doc/basics/route_syntax.md
+++ b/doc/basics/route_syntax.md
@@ -84,3 +84,8 @@ Routes are just data, so it's easy to create them programmatically:
 ;   ["/add-user" {:post {:interceptors [add-user]}}]
 ;   ["/add-order" {:post {:interceptors [add-order]}}])]
 ```
+
+### Encoding
+
+Reitit does not apply any encoding to your paths. If you need that, you must encode them yourself. 
+E.g., `/foo bar` should be `/foo%20bar`.
--- a/modules/reitit-core/src/reitit/core.cljc
+++ b/modules/reitit-core/src/reitit/core.cljc
@@ -112,17 +112,6 @@
 (defn find-names [routes _]
   (into [] (keep #(-> % second :name)) routes))
 
-(defn- encode-route
-  "URL-encode all non-parameter segments"
-  [[p m]]
-  (let [segments (str/split p #"/")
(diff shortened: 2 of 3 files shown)


KingMob:
@ikitommi Any thoughts?

ikitommi:
Sorry for taking so long time. 

Just tested with some 3rd party routers:

* `"/a b"`:
  * compojure -> fail
  * express -> fail
  * sinatra -> ok
* `"/100%best-product`
  * compojure -> ok
  * express -> ok
  * sinatra -> fail

from a browser:

* `/100%best-product` is sent as `/100%best-product`
* `/a b` is sent as `/a%20b`

... so none of those routers can handle these correctly.

To fix this for good, we should know how the clients send the uris. If this would be fixed, I should be pluggable - as the core router can be used in non-http routing too (routing in queuing system, which might allow all chars).

Are you actually facing this issue? if not, I would suggest not to fix this, just document it = 2, would work like compojure & express - and most of the routers out there.

KingMob:
Nah, not actually facing this issue. Decoding yes, encoding no.

I changed the PR to just add a note in the docs about not encoding paths. If you like it, cool, if not, no worries!

ikitommi:
Big thanks!

KingMob:
👍

PR #125, opened 2018-08-22
Frontend fixes

Deraen:
Related #77 

Though I'm not sure if we should just keep ignoring the fragment.


ikitommi:
LGTM

ikitommi:
If there is no known use case for tye fragment, let's not use it? Easy to add later if.

PR #128, opened 2018-08-25
balance missing parens

Bigsy:
The Swagger api example wasn't compiling as it was missing a paren, here is that missing paren in all its glory.


ikitommi:
Thanks!

PR #148, opened 2018-09-23
Generate options-endpoints for ring by default

ikitommi:
* `:options` requests are served for all routes by default with 200 OK to better support things like [CORS](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing)
  * the default handler is undocumented (for Swagger)
  * new router option `:reitit.ring/default-options-handler` to change this behavior. Setting `nil` disables this.


CHANGELOG.md
```diff
@@ -31,12 +31,34 @@
 ; #object[reitit.core$single_static_path_router]
 ```
 
+* `:options` requests are served for all routes by default with 200 OK to better support things like [CORS](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing)
+  * the default handler is undocumented (for Swagger)
```
miikka: I would maybe rephrase this as `is not documented in Swagger`. I get what you're saying, but it sounds like the feature is undocumented in the reitit docs.

doc/ring/ring.md
```diff
@@ -2,26 +2,63 @@
...
-(def app
-  (ring/ring-handler
-    (ring/router
-      ["/ping" handler])))
+(def router
+  (ring/router
+    ["/ping" {:get handler}]))
+```
+
+Match contains the ring-optimized `:result`:
```
miikka: What does ring-optimized mean?

doc/ring/ring.md
```diff
@@ -36,51 +73,73 @@ Applying the handler:
...
+;                              :handler #object[...]
+;                              :middleware []}}]]
 ```
 
-Note the compiled resuts as third element in the route vector.
-
 # Request-method based routing
 
-Handler are also looked under request-method keys: `:get`, `:head`, `:patch`, `:delete`, `:options`, `:post` or `:put`. Top-level handler is used if request-method based handler is not found.
+Handlers can be placed either to the top-level (all methods) or under a spesific method (`:get`, `:head`, `:patch`, `:delete`, `:options`, `:post`, `:put` or `:trace`). Top-level handler is used if request-method based handler is not found.
```
miikka: [nit] `spesific` -> `specific`

doc/ring/ring.md
```diff
@@ -36,51 +73,73 @@ Applying the handler:
...
+
+```clj
 (app {:request-method :get, :uri "/ping"})
 ; {:status 200, :body "ok"}
 
 (app {:request-method :put, :uri "/ping"})
 ; nil
 ```
 
+By default, `:options` is also supported (ree router options to change this):
```
miikka: `ree`? maybe `see`? Can you add a link?

doc/ring/ring.md
```diff
@@ -36,51 +73,73 @@ Applying the handler:
...
-    :path)
+    (r/match-by-name ::ping)
+    (r/match->path))
 ; "/ping"
 ```
 
 # Middleware
 
-Middleware can be added with a `:middleware` key, either to top-level or under `:request-method` submap. It's value should be a vector of any the following:
+Middleware can be mounted using a `:middleware` key - either to top-level or under `:request-method` submap. It's value should be a vector of `reitit.middleware/IntoMiddleware` values. These include:
```
miikka: `It's` -> `Its`. 

With "`:request-method` submap", do you mean under e.g. `:get` map? I think it'd be better to say "request method submap", then, because at first I thought `:request-method` is the concrete key for the map.

modules/reitit-ring/src/reitit/ring.cljc
```diff
@@ -57,20 +65,28 @@
...
                        :delete {:middleware [wrap-delete]
-                               :handler delete-user}}]])
-
-  See router options from [[reitit.core/router]] and [[reitit.middleware/router]]."
+                               :handler delete-user}}]])"
   ([data]
    (router data nil))
   ([data opts]
-   (let [opts (meta-merge {:coerce coerce-handler, :compile compile-result} opts)]
+   (let [opts (merge {:coerce coerce-handler
```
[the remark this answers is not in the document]
miikka: Fair enough.

PR #151, opened 2018-09-23
test for examples/ring-spec-swagger

johannesloetzsch:


examples/ring-spec-swagger/test/example/server_test.clj
```diff
@@ -0,0 +1,31 @@
+(ns example.server-test
+  (:require [clojure.test :refer :all]
+            [example.server :refer [app]]))
+
+(deftest example-server
+
+  (testing "GET"
+    (is (= (-> {:request-method :get :uri "/math/plus" :query-params {:x 20 :y 3}}
```
ikitommi: `:query-params` are received as string keys and string values from parameters mw. e.g. `{"x" "20", "y" "3"}`. To test that mw too, there could just be `:query-string` of `"x=30&y=3"` (as received from the adapter)

ikitommi:
Looks good. Depending if one wants to test the whole stack, the request should mimic what's coming from the ring adapter (json string body as inputstreams with correct content type, another test with edn body.). But it's ok to test just the parts. `:query-params` should be strings to see that the coercion actually works

johannesloetzsch:
> Looks good. Depending if one wants to test the whole stack, the request should mimic what's coming from the ring adapter (json string body as inputstreams with correct content type, another test with edn body.). But it's ok to test just the parts. `:query-params` should be strings to see that the coercion actually works

good point :) I think https://github.com/metosin/reitit/pull/151/commits/fb91499efdf62e46bf1bfc1c340c2a29e6fd4e59 does it

ikitommi:
Big thanks!!

PR #152, opened 2018-10-05
Fix reader conditional example

dancek:
The previous example only worked by accident (because there was an even number of map keys with conditional values).

{:get #?(:clj 1), :post #?(:clj 2)} gets read as {:get :post} in a non-clj context. The splicing reader conditional or a :cljs (or :default) value is needed to have the code work as intended. Here we change the docs to use the splicing version, as that is (imho) more beautiful than adding another branch.


ikitommi:
thx

Issue #89, opened 2018-05-22
create-resource-handler has hard-coded 404 response


ikitommi:
should be configurable. Maybe via route data / router options (so that setting it once would effect all places).

Setting `nil` is not good here, as it's in the branch where the route has matched -> returning `nil` would cause 406 (not-acceptable).

https://github.com/metosin/reitit/blob/master/modules/reitit-ring/src/reitit/ring.cljc#L116

PR #155, opened 2018-10-10
Custom handler for resource 404

valerauko:
## Issue
#89 

## Changes
Added optional `:not-found-handler` param to `create-resource-handler`.

It defaults to the same `constantly` as `create-default-handler` does.

## Tests
All current clj tests are green.

I didn't add tests for the new parameter -- should I? If so, to all the test branches (inside/outside router, root/path)?


ikitommi:
Thanks!

PR #157, opened 2018-10-16
Fix tests and reflection warnings

valerauko:
Made tests pass in `master`. I hope these changes are the right way.

* type hinted the exception bits that were causing reflection warnings
* flipped the validation tests that failed for string ke (was this behavior changed at some point?)


ikitommi:
Hi. The parameters were changes on haste to make `ds/opt` work as a key, see https://github.com/valerauko/reitit/commit/52eeea3c99efe684b366bb6accd56be9bb98e542#diff-ff98df9840ce78947438bb904a7f5643. I think it's still not right, as the parameters can be also instances of `clojure.spec.alpha/Spec`. I would change the Spec and test to verify it accepts a `s/keys` for example

PR #158, opened 2018-10-22
Add step to verify cljdoc.edn as part of CI

martinklepsch:
Hope this is useful in spotting discrepancies between Git state and `cljdoc.edn` :v: 

via https://github.com/cljdoc/cljdoc/issues/145


ikitommi:
Awesome, thanks!

Issue #92, opened 2018-05-28
Handle both trailing & missing backslashes on routes


ikitommi:
Both extra & missing backslashes should be handled somehow - fast & easy to use. Maybe new router option(s)?

Currently, for a router:

```clj
(r/router
  [["/dashboard" ::dashboard]
   ["/files/*" ::files]])
```
Both `/dashboard/` and `/files` fail

[httprouter](https://github.com/julienschmidt/httprouter#features) has some logic to handle these.

Currently, clients can either strip or add a training `/` if a match is not found, but we should do better.

ikitommi:
Related: https://www.scalescale.com/tips/nginx/nginx-remove-trailing-slash/

PR #159, opened 2018-10-28
Handle trailing slashes

valerauko:
If this is the right direction to go, I'll fix all the failing tests.

## Issue
#92 

## Changes
* add .ash_history and .m2 to .gitignore (dev convenience)
* ~make every route end in a trailing slash (/)~
* ~add a trailing slash to the path at lookup time if it doesn't end in one~

## To do
- [x] test cases
- [x] documentation


esuomi:
Just a suggestion but please
 - make this configurable so that it can be enabled/disable on demand
 - default to stripping ending slash, not adding it

When URI spec is considered, path is divided into segments as delimited by `/`. The segment is allowed to be empty meaning `example.com///` has a path with three _empty_ segments; semantically it represents a structure of `{nothing} -> {nothing} -> {nothing}`. This has implications on URI-based caching and performance of lower level routing such as what load balancers and proxy servers do and enforcing trailing slashes also can even pollute user's browser history in case of temporary redirects.

As far as semantics are considered, the only hard rule is that URI path should start with a slash in cases where authority (_spec lingo for user info, host and port_) is missing and that is the only by-the-spec enforceable slash in URI. But if authority is present, the path is allowed to be empty so even there's a caveat... :)

I do want to point out that I do believe this is an useful feature for a plethora of reasons. I've merely have had way too many experiences over the years where subpar frameworks/server software/misconfigurations have gone haywire due to mishandling of paths. Which is kind of incredible as this is all just about one single character...

For better details on how URIs should be handled I wholeheartedly recommend WIkipedia's [Uniform Resource Identifier](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) article.

.gitignore
```diff
@@ -12,3 +12,6 @@ pom.xml.asc
 /node_modules
 /_book
 figwheel_server.log
+
+/.m2
+/.ash_history
```
miikka: How do you end up with these in the repo in the first place?

valerauko:
All right, I'll try implementing both of your suggestions.

.gitignore
valerauko: I tend to dev inside docker with a command like
```
docker run -it -w /root -v "$(pwd):/root" -p 9001:9001 clojure:lein-alpine sh
```
Since it's mounted as a volume, when I run the scripts in [the guide](https://metosin.github.io/reitit/development.html) or `lein repl`, these flow back.

ikitommi:
Looking at httprouter docs:

> Stop caring about trailing slashes: Choose the URL style you like, the router automatically redirects the client if a trailing slash is missing or if there is one extra. Of course it only does so, if the new path has a handler. If you don't like it, you can turn off this behavior.

what if there was just a extra ring-handler that can be used in the default-branch, that would do a lookup without a trailing slash and if there would be a route, it would return a redirect?

esuomi:
@miikka Going by memory here so unfortunately low on specifics, but some examples from my personal past:
 - Varnish Cache when misconfigured generates unique cache entry for each variant of path
 - Play Framework at one point (like 5-6 years ago) at least used to have several issues related to slash handling
 - PHP world has a lot of misopinionated regular expressions annihilating slashes, esp. multiple slashes with empty segments.
 - Back in the day Apache Wicket's restful router would not allow empty segments at all. One popular car sales/auctioning website here in Finland used that and their URLs were forced to be of form `example.cars/Honda,Toyota/-/-/automatic/-/-/2007/-/` because they wanted to use the _search_ terms as path segments but couldn't use empty segments, which of course meant implementing dash handling for each search term sanitization separately. While this particular solution was IMHO actually kind of neat and definitely not a huge workload, the technical reason behind it was not that fun.
 - Can't remember by name which system it was, some slightly more esoteric Java Servlet based thing, but it actually filtered the query to be available only to the last path segment, which meant that `example.org/foo?bar=baz` would deliver `bar=baz` to path `/foo` but not `/foo/`, because the path `/foo` had no trailing query but the second empty segment did.

All in all this is a problem of converge towards form where all pieces in the full request/response handling cycle are configured to work in the same way. Which is why I asked for the configuration flag. As most configurations for doing similar things on other tools seem to rely on internal redirects/rewrites (well, that's my experience at least) it makes me believe _less is more_ which is why I asked for the default to be non-additive :) This is of course just a personal preference, as long as it's easy to configure it doesn't really matter one way or another.

valerauko:
Removing the trailing slash instead works as well and only one test failure instead of 108.

However, your suggestion of redirecting instead might be the way to go. From a SEO standpoint having equivalent routes with and without slash results in "duplicate content" which is generally undesirable.

How should that be implemented though? Having *both* routes and (based on an option) one redirecting to the other constantly?

ikitommi:
Pushed a small change into master so that the default-branch can also access the router. Agree on comments for not adding any implicit / magic into this. 

Did a small spike based my previous comment:

something like this in `reitit.ring`: 

```clj
(defn redirect-trailing-slash-handler []
  (let [maybe-redirect (fn maybe-redirect [request]
                         (let [uri (:uri request)]
                           (if (str/ends-with? uri "/")
                             (let [stripped (subs uri 0 (-> uri count dec))
                                   router (::r/router request)]
                               (if (r/match-by-path router stripped)
                                 {:status 307
                                  :headers {"Location" stripped}
                                  :body ""})))))]
    (fn
      ([request]
       (some-> request maybe-redirect))
      ([request respond _]
       (some-> request maybe-redirect respond)))))
```

would allow this to solved outside of the router, elegantly:

```clj
(def app
  (ring/ring-handler
    (ring/router
      ["/ping" (constantly {:status 200, :body "pong"})])
    (ring/redirect-trailing-slash-handler)))

(app {:request-method :get, :uri "/ping/"})
; {:status 307, :headers {"Location" "/ping"}, :body ""}

(app {:request-method :get, :uri "/ping"})
; {:status 200, :body "pong"}
```

valerauko:
Reverted the earlier changes and added the handler as you suggested with options.

If it's all right, I'll add test cases and documentation for it too.

.gitignore
Deraen: As those aren't really specific to this project, maybe you should ignore those on your global git config?

https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

valerauko: You're right, gonna revert

ikitommi:
This is really good & tests too.

PR #162, opened 2018-10-31
Support stripping multiple slashes

dancek:
One inspiration for stripping/adding slashes automatically in reitit was
https://github.com/julienschmidt/httprouter which can handle any number
of slashes. Make the implementation a bit better by stripping as many
slashes as there are.

This still doesn't support redirecting from a multiple-slash URI to a
single-slash URI. That could be handled as a special case, but in the
end I think the correct solution would be to convert both the router and
request URIs to a canonical form before matching. Then the question of
slashes is just a question of formatting.


ikitommi:
thanks!

PR #163, opened 2018-11-04
Perf tuning

ikitommi:
Faster path-parameters for the default case (of no need to decode)

* 2.5x faster
* 10% better perf in real-life (opensensors) test
* 15% better perf in real-life (github-api) test

### DIFF
--- a/modules/reitit-core/src/reitit/impl.cljc
+++ b/modules/reitit-core/src/reitit/impl.cljc
@@ -25,7 +25,7 @@
   [f coll]
   (reduce-kv
     (fn [coll k v]
-      (if-let [v' (f v)]
+      (if-some [v' (f v)]
         (assoc coll k v')
         coll))
     coll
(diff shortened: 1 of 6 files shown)


modules/reitit-core/src/reitit/impl.cljc
```diff
@@ -19,15 +19,16 @@
...
+(defn maybe-map-values
+  "Applies a function to every value of a map, updates the value if not nil.
   Also works on vectors. Maintains key for maps, order for vectors."
   [f coll]
   (reduce-kv
-    (fn [m k v]
-      (assoc m k (f v)))
-    (empty coll)
+    (fn [coll k v]
+      (if-let [v' (f v)]
```
miikka: Should be `if-some` to match docstring.

ikitommi: cool, didn't know such existed.

PR #164, opened 2018-11-05
Update docs for slash-handler

smahood:
Composing routes isn't explained until later, so anyone reading the docs from start to finish would not know how to use the default-handler and slash-handler together until they read the resource-handler section. Took me a bit of time to realize this so hopefully something like this would save a bit of work for other new users.


ikitommi:
Good catch, thanks!

PR #182, opened 2018-11-21
Add support for vector based specs in Swagger generation

hjhamala:
This allows data specs to be in vectors as well as maps. Before the change data spec [{:foo string?}] would throw exception when generating Swagger.


ikitommi:
I believe it's not only about swagger, but also coercion.

ikitommi:
thanks!

PR #197, opened 2018-12-26
Reitit pedestal

ikitommi:
first-class module for Pedestal.


ikitommi:
Merging this as it's just moving code around.

PR #202, opened 2019-01-09
minor fix: added missing dependency

johannesloetzsch:
ring-mock is needed to run `lein test`at  examples/ring-spec-swagger


miikka:
Thanks!

PR #206, opened 2019-01-17
Back linear-router with linear scan on segments 

ikitommi:
* less code
* 2x faster on simple tests
* 4x faster on opensensors

=> ~2-4x faster.

### DIFF
--- a/perf-test/clj/reitit/opensensors_perf_test.clj
+++ b/perf-test/clj/reitit/opensensors_perf_test.clj
@@ -13,7 +13,8 @@
 
             [io.pedestal.http.route.definition.table :as table]
             [io.pedestal.http.route.map-tree :as map-tree]
-            [io.pedestal.http.route.router :as pedestal]))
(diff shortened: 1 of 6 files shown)


perf-test/clj/reitit/opensensors_perf_test.clj
```diff
@@ -579,19 +581,23 @@
...
     ;;  4774ns
-    (b! "calfpath-unroll" calfpath-unroll-f)
+    #_(b! "calfpath-unroll" calfpath-unroll-f)
 
     ;;  2821ns
     (b! "pedestal" pedestal-f)
 
     ;;  4803ns
-    (b! "calfpath-macros" calfpath-macros-f)
+    #_(b! "calfpath-macros" calfpath-macros-f)
```
miikka: Why comment all these out?

ikitommi: uncommented.

PR #208, opened 2019-01-21
Rewrote `reitit.core/path-conflicting-routes` using transducers

green-coder:
It is easier to read, now.

### DIFF
--- a/modules/reitit-core/src/reitit/core.cljc
+++ b/modules/reitit-core/src/reitit/core.cljc
@@ -65,17 +65,17 @@
   (cond->> (->> (walk raw-routes opts) (map-data merge-data))
            coerce (into [] (keep #(coerce % opts)))))
 
-;; This whole function might be more efficient and easier to understand with transducers.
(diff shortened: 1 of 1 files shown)


modules/reitit-core/src/reitit/core.cljc
```diff
@@ -65,17 +65,17 @@
...
+  (let [conflicting-routes
+        (into {}
+              (comp (map-indexed (fn [index route]
+                                   [route (into #{}
+                                                (filter #(impl/conflicting-routes? route %))
+                                                (subvec routes (inc index)))]))
+                    (filter (comp seq second)))
+              routes)]
+    (when (seq conflicting-routes)
+      conflicting-routes)))
```
miikka: [golf] `(when (seq x) x)` is equivalent to `(not-empty x)`

green-coder: That was indeed the function that I was looking for! Thanks, I will add it.

miikka:
Thanks!

Issue #33, opened 2017-09-08
Optional middleware dependency resolution


ikitommi:
Middleware could have keys like `:produces` and `:requires` with common pseudo-values like `:parsed-body`, `:catch-exceptions` or project-spesific `::authenticated-user` etc. Middleware-chain could be automatically constructed based on these or if some dependency is missing, the router can't be constructed (with descriptive message).

Like [Anger-interceptor](https://github.com/oliyh/angel-interceptor) but for Middleware.

Not sure if this is a good feature thou.

If implemented, should work together with Macchiato's middleware resolution.

ikitommi:
Actually, a good way would be just to have the dependencies work on the request keys. E.g. `wrap-fornat` produces `[[:parameters :body]]`, while `gen-wrap-coerce-request` depends on it. No extra abstractions, there could be modes for the dependency resolvers:

1) check: just validate deps
2) organize: re-organize based on deps
3) complete: add missing deps (from the explicit registry)  if possible

ikitommi:
There is now `::middleware/tranform` to enable doing this.

PR #210, opened 2019-01-25
Middleware/interceptor dependency resolution algorithm

nilern:
Starting point for #33.


nilern:
I have a bad feeling about the 3-arity post-order. Maybe it should instead take an optional parameter map.

ikitommi:
how would the parameter map work? I would like to test with this, so merging into develop now.

ikitommi:
oh, it didn't work. https://circleci.com/gh/metosin/reitit/1604

ikitommi:
fixed

PR #212, opened 2019-01-28
Fix handling links inside contenteditable

joe-loco:
The default click behavior inside a container that is `contenteditable` is to not follow the link so that the link’s text stays selectable and editable. 

See [this jsfiddle](http://jsfiddle.net/gtewd276/1/) that shows a `div[contenteditable]` with a link inside it.

The current reitit behavior is to try to follow the link when it’s route is supported. This PR adds an exception to not follow them when they are nested inside a `contenteditable`.

### DIFF
--- a/modules/reitit-frontend/src/reitit/frontend/history.cljs
+++ b/modules/reitit-frontend/src/reitit/frontend/history.cljs
@@ -2,7 +2,8 @@
   (:require [reitit.core :as reitit]
             [reitit.core :as r]
             [reitit.frontend :as rf]
-            [goog.events :as gevents])
+            [goog.events :as gevents]
+            [goog.dom :as gdom])
   (:import goog.Uri))
 
 (defprotocol History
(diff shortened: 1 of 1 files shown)


modules/reitit-frontend/src/reitit/frontend/history.cljs
```diff
@@ -68,6 +68,7 @@
               (let [uri (.parse Uri (.-href el))]
                 (when (and (or (and (not (.hasScheme uri)) (not (.hasDomain uri)))
                                (= current-domain (.getDomain uri)))
+                           (not (.closest el "[contenteditable]"))
```
Deraen: Afaik `closest` is not supported in IE without polyfill. `goog.dom.getAncestor` should be better choice for compatibility: 
`(dom/getAncestor el (fn [node] (.isContentEditable node)))`. And because this check is probably not the fastest, it is probably best to move it to end the `and` so that if `and` is short-circuited, this check doesn't have to run.

joe-loco: Thanks! I just pushed your suggestions.

Deraen:
I managed to simplify (and optimize) this further, `isContentEditable` property is inherited so we don't need to check the parents, just the anchor element: 
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/isContentEditable
http://jsfiddle.net/2z3vhu0w/
https://github.com/metosin/reitit/commit/6123230f98f39ca02419828fd50a03db8e898a6a

Deraen:
And I also noticed we already uses `closest` just few lines up, so the Html5 History routing didn't work in IE, but I replaced that with alternative logic which works in IE: https://github.com/metosin/reitit/commit/cbff7d10bf400b111d10029971584424ab7caa13

joe-loco:
Great. Thanks!

PR #228, opened 2019-03-05
Update shared routes doc

alexanderjamesking:
Added missing require
Removed use of `or` and added the else branch to the `if` statement


ikitommi:
thanks!

PR #237, opened 2019-03-14
Announce that we use Break Versioning

miikka:
Let's make it official.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+# Reitit CHANGELOG
+
+We use [Break Versioning][breakver]. The version numbers follow a `<major>.<minor>.<patch>` scheme with the following intent:
+
+| Bump    | Intent                                                     |
+| ------- | ---------------------------------------------------------- |
+| `major` | Major breaking changes -- check the changelog for details. |
+| `minor` | Minor breaking changes -- check the changelog for details. |
+| `patch` | No breaking changes, ever!!                                |
+
+`-SNAPSHOT` versions are preview versions for upcoming releases.
+
+[breakver]: https://github.com/ptaoussanis/encore/blob/master/BREAK-VERSIONING.md
+
 ## 0.3.0-SNAPSHOT
 
 ### `reitit.core`
--- a/doc/development.md
+++ b/doc/development.md
@@ -25,6 +25,10 @@ gitbook serve
 
 ## To bump up version:
 
+We use [Break Versioning][breakver]. Remember our promise: patch-level bumps never include breaking changes!
+
+[breakver]: https://github.com/ptaoussanis/encore/blob/master/BREAK-VERSIONING.md
+
 ```bash
 # new version
 ./scripts/set-version "1.0.0"


esuomi:
This is horrible! :(

Break versioning is a massive misunderstanding of software version parity in relation to release version parity - that is, what "version" means for marketing and sales is not the same as what it semantically carries to developers. Release version is another layer of versioning and can be a composite of multiple software versions; as good examples Windows 7 ("release version", in a sense) is in fact software version 6.1, build 7601.

To be specific about the issues as highlighted by the [break versioning doc](https://github.com/ptaoussanis/encore/blob/master/BREAK-VERSIONING.md),

 1. Sometimes length is required to be comprehensive. And besides Semantic Versioning isn't long, it is just written in way that makes sure all corners are taken into consideration.
 2. Software breaks are common etc. but practically the severity and scope of the breaking change have nothing to do with the fact that *from userland perspective they are equal*; all breaking changes must be analyzed properly to understand the scope of required modifications if update is to be made and in larger contexts if the refinement of the change is larger than the effort vs benefit, the library won't updated. Labeling all changes as "breaking" in fact prevents doing any updates at all in stricter environments and is very counter-productive as it will lead up to larger version fragmentation of the library's usage.
 3. Market/psychological pressure I already described above in the first paragraph.
 4. Pre-1.0.0 anarchy is a matter of social structure within the team and the magical commitment to 1.0 being some kind of holy cow. While Semantic Versioning recommends starting from 0.1.0 that's all there is to it - if your second commit is a breaking change, roll the version to 1.0.0 and just go with it. No one really cares and you can always market package the software separately.

There is also bigger social issue here. Most people assume significance in version numbers on such level that the change effect parity between 2.0.0 and 3.0.0 is same as 6.0.0 and 7.0.0 - that is, the +1 increment on either level are equal. Allowing for subjective definition of increment scope (what's major for you is minor for me and vice versa) means that you'll either have to write a document about how the increment qualification is assessed or allow total anarchy (post-v1.0? so reverse of reason no. 4) and that will definitely make Break Versioning spec longer than SemVer.

---

That said, this is already merged in and obviously in larger context you prefer Break Versioning and are using it elsewhere so _*BLEH*_, it is what it is. Now I can at least in future point out to this post when someone else asks me about this. I'll take my ranty self elsewhere now :)

miikka:
@esuomi, if you want to discuss Peter Taoussanis's critique of Semantic Versioning, you should discuss it with him. If you want to discuss Reitit versioning policy, please make a concrete proposal, because frankly I don't really get what you're trying to say.

esuomi:
@miikka Well, this was more of an emotional reaction to this announcement driven directly from GitHub notifications to first available input box... :)

I'm totally fine with _not_ having a discussion about this, it is your decision after all. While I would prefer SemVer being used for Reitit as well, I don't feel it's a sensible proposal in _announcement_ of moving to another versioning scheme as that'd be quite rude of me, even.

miikka:
Yeah, what I'd like to do next is to wait a while and see how having this explicit policy works. What I hope to happen:

* When merging patches, we will think about they relate to this policy and how the version number should be incremented. Relevant questions: Will it be easy to decide on the version increment?  Do people agree about it? Will we ever use patch-level bumps?
* When users (incl. ourselves) are upgrading Reitit, if they end up being confused or frustrated by the versioning scheme, I hope that they will ask help or complain.

Once we have some experience with this – say, in 6-12 months – I think we could revisit this question. If everything has been smooth, though, a new proposal would have to be *really* convincing.

Issue #238, opened 2019-03-15
Handle trailing slashes in frontend module


Deraen:
Similar to ring module, frontend should provide options to handle trailing slashes in routes:

https://github.com/metosin/reitit/blob/master/modules/reitit-ring/src/reitit/ring.cljc#L111-L139

Ring module only does HTTP redirect, but for frontend it would probably make sense to provide options to do redirect (`replaceState` probably) or to just return the same match from path with or without `/`.

sudorock:
Hello, can we expect this change to be in master anytime soon?

aiba:
This would be handy.

PR #239, opened 2019-03-15
Implement #238, handle trailing slashes in frontend match-by-path

Deraen:
Does adding option to router make sense? Are there other ways to control this logic?

### DIFF
--- a/modules/reitit-frontend/src/reitit/frontend.cljs
+++ b/modules/reitit-frontend/src/reitit/frontend.cljs
@@ -1,5 +1,6 @@
 (ns reitit.frontend
   (:require [clojure.set :as set]
+            [clojure.string :as str]
             [reitit.coercion :as coercion]
             [reitit.coercion :as rc]
             [reitit.core :as r])
@@ -16,10 +17,18 @@
 
 (defn match-by-path
   "Given routing tree and current path, return match with possibly
-  coerced parameters. Return nil if no match found."
+  coerced parameters. Returns nil if no match found."
   [router path]
-  (let [uri (.parse Uri path)]
-    (if-let [match (r/match-by-path router (.getPath uri))]
+  (let [uri (.parse Uri path)
+        path (.getPath uri)]
+    (if-let [match (or (r/match-by-path router path)
+                       (if-let [trailing-slash-handling (:trailing-slash-handling (r/options router))]
+                         (if (str/ends-with? path "/")
+                           (if (not= trailing-slash-handling :add)
+                             (r/match-by-path router (subs path 0 (dec (count path)))))
+                           (if (not= trailing-slash-handling :remove)
+                             (r/match-by-path router (str path "/"))))))]
+      ;; User can update browser location in on-navigate call using replace-state
       (let [q (query-params uri)
             match (assoc match :query-params q)
             ;; Return uncoerced values if coercion is not enabled - so
@@ -40,7 +49,14 @@
 
 (defn router
   "Create a `reitit.core.router` from raw route data and optionally an options map.
-  Enables request coercion. See [[reitit.core/router]] for details on options."
+  Enables request coercion. See [[reitit.core/router]] for details on options.
+
+  Additional options:
+
+  | key          | description |
+  | -------------|-------------|
+  | :trailing-slash-handling | TODO |
+  "
   ([raw-routes]
    (router raw-routes {}))
   ([raw-routes opts]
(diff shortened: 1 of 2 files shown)


modules/reitit-frontend/src/reitit/frontend.cljs
```diff
@@ -16,10 +17,20 @@
...
 (defn match-by-path
   "Given routing tree and current path, return match with possibly
-  coerced parameters. Return nil if no match found."
+  coerced parameters. Returns nil if no match found."
   [router path]
-  (let [uri (.parse Uri path)]
-    (if-let [match (r/match-by-path router (.getPath uri))]
+  (let [uri (.parse Uri path)
+        path (.getPath uri)]
+    (if-let [match (or (r/match-by-path router path)
```
ikitommi: This will make the lookup much slower as there is an extra hash-lookup for each routing call. Would be better to move the check into router creation time and return a reified `r.c/Router` with different code for the different cases.

modules/reitit-frontend/src/reitit/frontend.cljs
```diff
@@ -40,7 +51,14 @@
...
 (defn router
   "Create a `reitit.core.router` from raw route data and optionally an options map.
-  Enables request coercion. See [[reitit.core/router]] for details on options."
+  Enables request coercion. See [[reitit.core/router]] for details on options.
+
+  Additional options:
+
+  | key          | description |
+  | -------------|-------------|
+  | :trailing-slash-handling | TODO |
```
ikitommi: could be just `:trailing-slash` (or `:trailing-slash-method` (the ring-version has `:method`)).

ikitommi: TODO -> https://github.com/metosin/reitit/blob/master/modules/reitit-ring/src/reitit/ring.cljc#L117-L119

modules/reitit-frontend/src/reitit/frontend.cljs
ikitommi: Still think this would be cleaner to be implemented as a reified Proxy, but that could be done later too.

Deraen:
Now implemented as custom Router, which wraps the parent router.

TODO:
- [ ] Move to reitit-core
- [ ] Migrate reitit-ring to use this.
- [ ] Decide on API, do we add option on some basic routers? Maybe even `reitit.core/router`, or should user wrap another router in trailing-slash-router?
- [ ] Check performance, alternative implementation would be to modify routes to match additional / missing slashes in Router creation time.

Deraen:
I remembered some reason why it might not be necessary or even good idea to move this to core:

Frontend and backend have quite a different solutions on what to do in these cases: this frontend solution just returns the match as if the original route was matched. In backend ring handler will return redirect response.

orestis:
I can't talk about the implementation details (moving to core or not), but my 2c on (my) desired behaviour in the frontend:

- Generally speaking, would like to have a "canonical" representation of the route. It's probably application specific about whether this is with, or without a trailing slash. This is something we can control today by using `""` or `"/"` in a route definition.
- Any non-canonical but equivalent paths are coming from either users typing directly into the URL bar (e.g. stripping off a path segment) or external applications linking in (where someone types the URL into an input field). In both cases, we can't control what the link looks like.
- Would like the frontend to silently accept the equivalent path, and route to the canonical one. Sometimes we might want to change the visible URL to reflect the canonical path, but sometimes that's not necessary.

Thank you for your work on this!

orestis:
Not sure if useful for anyone, but here's a small snippet that seems to do the trick:

`route!` and `navigation-fail` are application-specific handlers. The `try-without-slash` is run upon a match fail, to see if we can recover. 
```
(defn- try-without-slash [path router]
  (if (str/ends-with? path "/")
    (let [non-slash-path (subs path 0 (dec (count path)))]
      (if-let [match (rf/match-by-path router non-slash-path)]
        (do
          (.replaceState js/window.history nil "" non-slash-path)
          (route! match))
        (navigation-fail non-slash-path)))
    (navigation-fail path)))
```

modules/reitit-frontend/src/reitit/frontend.cljs
```diff
@@ -14,12 +15,38 @@
...
+        (r/route-names parent))
+      (match-by-path [_ path]
+        (or (r/match-by-path parent path)
+            (if (str/ends-with? path "/")
+              (if (not= method :add)
+                (r/match-by-path parent (subs path 0 (dec (count path)))))
+              (if (not= method :remove)
+                (r/match-by-path parent (str path "/"))))))
+      (match-by-name [_ name]
+        (r/match-by-name parent name)))
```
moonbrv, line 40: also need implementation 
```
(match-by-name [_ name path-params]
        (r/match-by-name parent name path-params))
```

Issue #246, opened 2019-03-22
Swagger spec generation


jameslintaylor:
Currently, If one wishes to leverage the swagger 2.0 spec generation implemented in `reitit.swagger`, it _must_ be done via `reitit.swagger/create-swagger-handler`.

This is fine when planning to serve the spec at an endpoint, but makes generating a spec for a set of compiled routes tedious otherwise (ie. mocking a request to the handler).

jameslintaylor:
See https://github.com/metosin/reitit/pull/247

ikitommi:
Ho, and thanks for the detailed issue and the PR. Just curious, what is the use case for generating the spec outside of the request cycle? Will you still have the `create-swagger-handler` mounted?

Will comment the PR separately.

jameslintaylor:
The specific use case was generating a swagger 2.0 spec intended to be saved to the filesystem, such that it could be consumed later via [swagger-codegen](https://github.com/swagger-api/swagger-codegen), [bootprint](https://github.com/bootprint/bootprint-openapi#overview), or similar.

PR #247, opened 2019-03-22
Extract swagger spec generation logic from `create-swagger-handler`

jameslintaylor:
Fixes https://github.com/metosin/reitit/issues/246

This extracts the logic for doing the swagger 2.0 spec creation from the `reitit.swagger/create-swagger-handler` as a public api `(reitit.swagger/swagger-spec compiled-routes swagger-data)`.

### DIFF
--- a/modules/reitit-swagger/src/reitit/swagger.cljc
+++ b/modules/reitit-swagger/src/reitit/swagger.cljc
@@ -68,40 +68,55 @@
 (defn- swagger-path [path]
   (-> path trie/normalize (str/replace #"\{\*" "{")))
 
-(defn create-swagger-handler []
-  "Create a ring handler to emit swagger spec. Collects all routes from router which have
-  an intersecting `[:swagger :id]` and which are not marked with `:no-doc` route data."
+(defn swagger-spec
+
+  "Return a map representative of the OpenAPI/swagger 2.0 for
+  `compiled-routes`, merging the top level `swagger-data`.
+
+  Usage:
+    (swagger-spec {:info {:title \"The Bestest API\"}} compiled-routes)"
+
+  [swagger-data compiled-routes]
+  (let [->set (fn [x] (if (or (set? x) (sequential? x)) (set x) (conj #{} x)))
+        ids (->set (:id swagger-data ::default))
+        strip-top-level-keys #(dissoc % :id :info :host :basePath :definitions :securityDefinitions)
+        strip-endpoint-keys #(dissoc % :id :parameters :responses :summary :description)
+        swagger (->> (strip-endpoint-keys swagger-data)
+                     (merge {:swagger "2.0"
+                             :x-id ids}))
+        accept-route (fn [route]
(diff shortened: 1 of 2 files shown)


ikitommi:
Few quick comments:

* the swagger api can be scoped to just certain routes using `[:swagger :id]` route data. The proposed new function takes `compiled-routes`, is the idea that the caller knows which routes to include?
* There multiple swagger apis within a router, as described [in the docs](https://cljdoc.org/d/metosin/reitit/0.3.1/doc/ring/swagger-support#multiple-swagger-apis)

What if there was a function that takes a ring-handler (or a ring-router) and optionally the swagger-idand would return the docs for that? It would take care of collecting the correct routes and would support the multiple api-case. 

Something like:

```clj
(def app
  (ring/ring-handler
    (ring/router
      [["/default" handler]
       ["/one"
        {:swagger {:id ::one}
         :get handler}]
       ["/both"
        {:swagger {:id #{::one ::two}}
         :get handler}]] ] ) ) )

(swagger-spec app)
(swagger-spec app ::swagger/default)
;; "/default"

(swagger-spec app ::one)
;; "/one" & "/both"

(swagger-spec app ::api2)
;; "/both"

(swagger-spec app ::swagger/default ::api2)
;; "/default" & "/both"
```

jameslintaylor:
Sorry for the late reply - I believe that `(swagger-spec swagger-data compiled-routes)`, as it is defined, should already support this. To mirror your examples:

```clojure
(def routes 
 [["/default" handler]
  ["/one"
   {:swagger {:id ::one}
    :get handler}]
  ["/both"
   {:swagger {:id #{::one ::two}}
    :get handler}]]])

(def router (ring/router routes))
(def app (ring/ring-handler router))
(def compiled-routes (reitit/compiled-routes router))

(swagger-spec {} compiled-routes)
(swagger-spec {:id ::swagger/default} compiled-routes)
;; "/default"

(swagger-spec {:id ::one} compiled-routes)
;; "/one" & "/both"

(swagger-spec {:id ::two} compiled-routes)
;; "/both"

(swagger-spec {:id #{::swagger/default ::two}} compiled-routes)
;; "/default" & "/both"
```

jameslintaylor:
> What if there was a function that takes a ring-handler (or a ring-router) and optionally the swagger-idand would return the docs for that? It would take care of collecting the correct routes and would support the multiple api-case.

The main reason I decided to pass `swagger-data` in `(swagger-spec swagger-data compiled-routes)` is such that the consumer could specify more than just `:id`. For instance, top level api title/descriptions available via `:title`/ `:description`  ( typically defined as data on the route mounting `(create-swagger-handler)`) could be specified:

```clojure
(swagger-spec {:title "My API"
               :description "Describe my API"
               :id ::one}
              compiled-routes)
```

Now that you mention it, passing the router itself might make these examples cleaner (we wouldn't need to do the `(comp reitit/compiled-routes ring/handler)` dance)... but I'm not sure if that adds much value.

PR #251, opened 2019-04-02
Remove unused compojure dependencies from frontend examples

aarongroom:
These don't appear to be needed and might cause users to introduce
dependencies into their projects if they use as templates in their own
projects.


miikka:
Well-spotted, thanks!

Issue #250, opened 2019-04-02
Support Sieppari-style error handlers with pedestal


ikitommi:
the `Executor` in `reitit-pedestal` could check if the mounted interceptors have 1-arity error handler. If they do, it should wrap the `:error` into a 2->1 arity converter so that ALL Sieppari-style interceptors would work oob with Pedestal.

Clojure doesn't support asking function arity at runtime (megabummer!), but we can hack around it: https://stackoverflow.com/questions/1696693/clojure-how-to-find-out-the-arity-of-function-at-runtime

vielmath:
I'm currently facing the issue and it seems to be a good fix.

vielmath:
I think this issue should be closed, #254 has been merged.
Thanks.

ikitommi:
I'll add some tests and cut out a new version with this (and other small fixes) in.

ikitommi:
This is shipped in `[metosin/reitit "0.3.2"]`

PR #254, opened 2019-04-04
Support Sieppari-style error handlers with pedestal

vielmath:
My suggestion concerning #250 

The reitit-pedestal Executor checks if the mounted interceptors have
1-arity error handler. If so, they are wrapped into a 2 arrity function.

Now it's done, I'm not sure about the need to transparently allow Sieppari-style error interceptors.
I might be better to just expose the `wrap-error-arity-2->1` helper so the user can use it if he knows what he is doing...

### DIFF
--- a/modules/reitit-pedestal/src/reitit/pedestal.clj
+++ b/modules/reitit-pedestal/src/reitit/pedestal.clj
@@ -6,6 +6,29 @@
             [reitit.http])
   (:import (reitit.interceptor Executor)))
 
+(defn- arity [f]
+  (->> (class f)
+       .getDeclaredMethods
+       (filter #(= "invoke" (.getName %)))
+       first
+       .getParameterTypes
+       alength))
+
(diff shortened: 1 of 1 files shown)


ikitommi:
Had to sleep on this. My 2 cents fo on having an automatic bridging on by default. Why?

* just having a helper will cause problems on most/all of the users and the failure will happen at runtime - and only on case of exception => not good
   * the solution for a user to fix this with the helper is either: 
      a) wrap manually all known interceptors with `:error` with the helper fn => need to check all interceptors currently AND in the future => potentially more errors in the future
      b) add a script / option to automatically to this for all future cases => why nyt on by default?

There could be an indication that bridging is applied, but not sure if that has any value to the users.

Will ask a third opinion. ping @miikka .

Also, there are no tests for this, but also none for the `reitit-pedestal` module, we should add some, but can be done later.

ikitommi:
Thanks!

PR #256, opened 2019-04-09
Fix: Ignore anchors with any target value except empty or _self

alexanderjamesking:
We should probably ignore `_parent` and `_top` as well?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

### DIFF
--- a/modules/reitit-frontend/src/reitit/frontend/history.cljs
+++ b/modules/reitit-frontend/src/reitit/frontend/history.cljs
@@ -84,7 +84,7 @@
                            (not (.-ctrlKey e))
                            (not (.-metaKey e))
                            (not (.-shiftKey e))
-                           (not (contains? #{"_blank" "self"} (.getAttribute el "target")))
+                           (not (contains? #{"_blank" "_self"} (.getAttribute el "target")))
                            ;; Left button
                            (= 0 (.-button e))
                            ;; isContentEditable property is inherited from parents,
--- a/modules/reitit-frontend/src/reitit/frontend/history.cljs
+++ b/modules/reitit-frontend/src/reitit/frontend/history.cljs
@@ -84,7 +84,8 @@
                            (not (.-ctrlKey e))
                            (not (.-metaKey e))
                            (not (.-shiftKey e))
-                           (not (contains? #{"_blank" "_self"} (.getAttribute el "target")))
+                           (or (not (.hasAttribute el "target"))
+                               (contains? #{"" "_self"} (.getAttribute el "target")))
                            ;; Left button
                            (= 0 (.-button e))
                            ;; isContentEditable property is inherited from parents,


Deraen:
Reactor-router ignores anchors with any `target` except `_self`. I guess react-router reasoning is that because `_self` means that link should be opened in current context, it should be handled using History router if other checks are true: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L21

alexanderjamesking:
Interesting, I was using it to bypass the router and force a hard reload of the page. `self` should probably be removed anyway then as it seems a strange thing to check for as it's not actually a valid option for `target`. 

I was using pushy previously which checked for `_blank` and `_self` [here](https://github.com/kibu-australia/pushy/blob/ce02504d811fe7d63f7438e3dda34d6cd519a970/src/pushy/core.cljs#L112). It was a convenient way to bypass the router but perhaps it's abusing what it was actually intended for.

I guess I can work around it with an `on-click` event and trigger the redirect manually, it's probably right on the edge of the scope of what reitit should be dealing with anyway, but it'd be nice to have something on a link to make it easier to bypass the router for when it's required.

Deraen:
Accountant also uses the same logic as react-router: https://github.com/venantius/accountant/commit/81140a4c35be7b2c8a1a3c6d0b20145d38282e66

I think the correct solution here is to match react-router and Accountant, and handle links with `_self` but ignore anchors with any target value except empty or `self_`:

```
(contains? #{"" "_self"} (.getAttribute el "target"))
```

Note, without `not`

Deraen:
Hmm. Pushy code definitely looks familiar, I was wondering where I got this check from.

What is your use case for triggering hard reload? Looking at react-router and `target` docs, I think using `self_` for this purpose is probably not very good idea.

Though react-router is a bit different, because their code only runs for special `Link` components, not on any anchor click.

alexanderjamesking:
Our use case is down to some legacy code that changes the context of the user where we have server side code modifying the cookie, ideally we wouldn't hard reload at all but that requires a bit more work, so it's really just a short term hack to force the reload.

The only solution I can think of for now would be to add something to the dom that reitit reads and decides to ignore, but I wouldn't recommend that anyway. I can work around it for now until I actually fix the real problem in our codebase, but I'll submit a change to bring things inline with react-router and Accountant.

Thanks for investigating.

Deraen:
Simplest solution would be if the legacy route would be outside your routing tree, because only anchors with matching href are handled.

We could also check the route data for matched route, but I don't like this idea either.

alexanderjamesking:
Another option would be to allow `ignore-anchor-click` to be passed in when the router is constructed, we could default to the current behaviour if no fn is passed. It's a nice option as it allows users to customise this fn if needed and enforces no contract on the structure of the dom.

alexanderjamesking:
No longer required as this is handled in #259

PR #258, opened 2019-04-10
Support html5 links inside Shadow DOM

fraxu:
We are using web components so we needed to get html5 links inside shadow dom. So we needed to get correct target element from MouseEvent's composedPath() instead of just using the target property. 

We also added a test that clicks links to test this. Test will fail with "Some of your tests did a full page reload!" if test fails.

If you want any changes to this, I'm happy to help!

### DIFF
--- a/modules/reitit-frontend/src/reitit/frontend/history.cljs
+++ b/modules/reitit-frontend/src/reitit/frontend/history.cljs
@@ -60,8 +60,8 @@
           (recur (.-parentNode el)))))))
 
 (defn- event-target [event]
-  ;; Read event's target from composed path to get shadow dom working,
-  ;; fallback to target property if not available
+  "Read event's target from composed path to get shadow dom working,
+   fallback to target property if not available"
   (let [original-event (.-event_ event)]
     (if (exists? (.-composedPath original-event))
       (first (.composedPath original-event))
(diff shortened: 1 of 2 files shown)


modules/reitit-frontend/src/reitit/frontend/history.cljs
```diff
@@ -59,6 +59,14 @@
           el
           (recur (.-parentNode el)))))))
 
+(defn- event-target [event]
+  ;; Read event's target from composed path to get shadow dom working,
+  ;; fallback to target property if not available
```
alexanderjamesking: This would be better in a doc-string instead of a comment

fraxu: Thanks for the input. I changed this to doc-string.

modules/reitit-frontend/src/reitit/frontend/history.cljs
```diff
@@ -59,6 +59,14 @@
           el
           (recur (.-parentNode el)))))))
 
+(defn- event-target [event]
+  "Read event's target from composed path to get shadow dom working,
+   fallback to target property if not available"
+  (let [original-event (.-event_ event)]
```
Deraen: Use `.getBrowserEvent` method instead of the internal `event_` property.

Deraen:
I will need to read a bit to understand this better, I don't really have any experience with Shadow DOM. Luckily the change is quite small, I'll try to find time to check this properly tomorrow.

modules/reitit-frontend/src/reitit/frontend/history.cljs
fraxu: Changed

PR #259, opened 2019-04-15
Ignore anchor click fn

Deraen:
Follow up to #256


alexanderjamesking:
Thanks for your work on this, the changes look good to be. The lines haven't changed as part of this PR but I think it's worth renaming the [ignore-anchor-click](https://github.com/metosin/reitit/blob/f71b3c0a3bf2632edbe74e40be543b94ed06f6ec/modules/reitit-frontend/src/reitit/frontend/history.cljs#L79)  fn to something like `click-handler` or `on-click`, as it's current name does not represent all of what the function is actually responsible for.

If users require further customisation (e.g. modifier key + click) then perhaps it's better for them to implement their own Html5History record that implements the Reitit History protocol.

Deraen:
TODO: check option naming

Deraen:
I refactored this a bit so that the function is only the predicate, so it will be easier to customize logic completely. I think the option name is probably okay now.

doc/frontend/browser.md
```diff
@@ -18,6 +18,31 @@ request to the server. This means the URL will look normal, but the downside is
 that the server must respond to all routes with correct file (`index.html`).
 Check examples for simple Ring handler example.
 
+### Anchor click handling
+
+HTML5 History router will handle click events on anchors where the href
+matches the route tree (and other [rules](../../modules/reitit-frontend/src/reitit/frontend/history.cljs#L84-L98)).
+If you have need to control this logic, for example to handle some
+anchor clicks where the href matches route tree normally (i.e. browser load)
+you can provide `:ignore-anchor-click` function to add your own logic to
```
miikka:
```suggestion
you can provide `:ignore-anchor-click?` function to add your own logic to
```

ikitommi:
How could we finalize this? About to release 0.3.2, guess there is some naming thing related to this?

Deraen:
I think the naming is fine so this should be OK to merge.

PR #260, opened 2019-04-20
Frontend examples

vharmain:
# Frontend examples

## React-router examples

Examples how to implement simplified versions of following `react-router` features using Reitit frontend.

**frontend-links** folder
- https://reacttraining.com/react-router/web/api/Link
- https://reacttraining.com/react-router/web/api/NavLink
- https://reacttraining.com/react-router/web/api/Redirect

**frontend-prompt** folder
- https://reacttraining.com/react-router/web/api/Prompt

## Re-frame example

**frontend-re-frame** folder
Example how to integrate reitit-frontend with re-frame.

Any feedback, improvements and other suggestions are highly welcome! I have no prior experience with `react-router` and I'm not sure if the selected examples are the most useful ones. I'm happy to fix these and implement more examples if they seem useful.

## Bonus bugfix
Fixes a minor bug in `reitit.core` causing redundant `?` in path if given empty query params `{}`.


examples/frontend-links/src/frontend/core.cljs
```diff
@@ -0,0 +1,143 @@
...
+  "If `push` is truthy, previous page will be left in history."
+  [{:keys [to path-params query-params push]}]
+  (if push
+    (rfe/push-state to path-params query-params)
+    (rfe/replace-state to path-params query-params)))
+
+(defn Redirect
+  "Component that only causes a redirect side-effect."
+  [props]
+  (redirect! props)
```
Deraen: Can't say I much like the idea of Component that only causes side-effects, but for sake of example, the implementation should match React-router more closely:

```cljs
(defn Redirect [props]
  (r/create-class
    {:component-did-mount (fn [this] (redirect! (r/props this)))
     :component-did-update (fn [this [_ prev-props]]
                             (if (not= (r/props this) prev-props)
                               (redirect! (r/props this))))
     :render (fn [this] nil)})

File: examples/frontend-links/src/frontend/core.cljs [hunk 2]
```diff
@@ -0,0 +1,143 @@
...
+
+(defonce current-match (r/atom nil))
+
+(defn- resolve-href
+  [to path-params query-params]
+  (if (keyword? to)
+    (rfe/href to path-params query-params)
+    (let [match  (rf/match-by-path router to)
+          route  (-> match :data :name)
+          params (not-empty (or path-params (:path-params match)))
```
Deraen: These `not-empty` calls are probably unnecessary?

examples/frontend-re-frame/src/cljs/frontend_re_frame/core.cljs
```diff
@@ -0,0 +1,157 @@
...
+       :stop  (fn [& params] (js/console.log "Leaving sub-page 1"))}]}]
+   ["sub-page2"
+    {:name      ::sub-page2
+     :view      sub-page2
+     :link-text "Sub-page 2"
+     :controllers
+     [{:start (fn [& params] (js/console.log "Entering sub-page 2"))
+       :stop  (fn [& params] (js/console.log "Leaving sub-page 2"))}]}]])
+
+(defn on-navigate [new-match]
```
Deraen: It is best to not use `subscribe` outside render or other subscriptions, it will create RAtoms which will be only used once etc.
I think it is better to handle this old-match, apply-controllers etc. in `::navigated` event handler, where it  is easy to read the old match from app-db etc.

examples/frontend-re-frame/src/cljs/frontend_re_frame/core.cljs
```diff
@@ -0,0 +1,157 @@
...
+
+(defn init-routes! []
+  (js/console.log "initializing routes")
+  (rfe/start!
+   router
+   on-navigate
+   {:use-fragment true}))
+
+(defn nav [{:keys [router current-route]}]
+  (into
```
Deraen: It is best to use `(into [...] seq)` only when necessary, when items don't have keys and/or the seq is static. Here route name can be used as unique key, so you can just move the `for` loop inside `[:ul ]` and add `:key route-name` to `:li`.

examples/frontend-links/src/frontend/core.cljs
vharmain: Thanks for the review!

I did change the example to use this implementation. However I'm curious about in what situation does `:component-did-update` get called? I'd expect that `redirect!` on mount would take the user elsewhere and make update scenario redundant?

examples/frontend-links/src/frontend/core.cljs
vharmain: I added them because empty query params `{}` cause redundant `?` in href. Maybe it could be fixed in `rfe/href`?

examples/frontend-re-frame/src/cljs/frontend_re_frame/core.cljs
vharmain: Good point! 👍

examples/frontend-re-frame/src/cljs/frontend_re_frame/core.cljs
vharmain: I seem to have a bad habit of overusing `into`. I changed it according to your comment.

examples/frontend-links/src/frontend/core.cljs
Deraen: The problems seems to be in `reitit.core/match->path`: https://github.com/metosin/reitit/blob/master/modules/reitit-core/src/reitit/core.cljc#L70

You could fix it in this PR, or open new issue for this problem & remove the workaround in the example.

vharmain: I decided to fix it and added the test case. I weighted wether I'd use `not-empty` or `seq`. Both worked but I decided to use `not-empty` in case someone in the future still expects to get a map instead of vectors downstream.

ikitommi:
Awesome, thanks!

PR #269, opened 2019-05-11
Polish for 0.3.2

ikitommi:
* tests for new pedestal arity-checker
* boost examples to enable easily the spec validation


ikitommi:
just more tests + updated examples, merging this.

PR #275, opened 2019-05-20
Properly access ::s/problems key in coercion-spec

kwrooijen:
## Problem

When making an API request with a payload that doesn't conform to the defined spec, I receive a `400` error, but the `problems` key contains an empty vector instead of a problem description.

## Solution

It seems that we're trying to access the `::s/problems` key from `error`, however this is wrapped within another `:problems` key. When accessing the nested key the issue is resolved.

## Notes

I'm not sure if this is the correct solution, maybe the key shouldn't be nested in the first place?

### DIFF
--- a/modules/reitit-spec/src/reitit/coercion/spec.cljc
+++ b/modules/reitit-spec/src/reitit/coercion/spec.cljc
@@ -110,7 +110,7 @@
       (into-spec model name))
     (-open-model [_ spec] spec)
     (-encode-error [_ error]
-      (let [problems (::s/problems error)]
+      (let [problems (-> error :problems ::s/problems)]
         (-> error
             (update :spec (comp str s/form))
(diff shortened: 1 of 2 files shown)


ikitommi:
Thanks! Could you also add a test assertion for this that it won't regress later. I think the right place is around https://github.com/metosin/reitit/blob/master/test/cljc/reitit/ring_coercion_test.cljc#L99

ikitommi:
Thanks!

kwrooijen:
Not a very comprehensive test, but it does cover the issue at hand. 

Do you by any chance have an ETA for when this is getting released in the new version? We can use a git checkout in the meanwhile so there's certainly no rush.

Thank you for your rapid responses!

ikitommi:
`[metosin/reitit "0.3.4"]`

PR #279, opened 2019-05-22
FIX: path-parameters concurrency

ikitommi:


test/cljc/reitit/ring_test.cljc
```diff
@@ -577,3 +578,19 @@
        (fn [{:keys [::r/router]} _ _]
          (is router)))
      {} ::respond ::raise)))
+
+#?(:clj
+   (deftest invalid-path-parameters-parsing-concurrent-requests-277-test
+     (testing "in enought concurrent system, path-parameters can bleed"
```
Deraen: concurrent enough

miikka:
This fix is okay but I'd prefer the fix suggested by @viesti: make `Match.params` final and create a new Match instead of doing this in `WildMatcher.match`:
https://github.com/metosin/reitit/blob/60ee39bd53e21842b8d68fab2f1e88251f13bf3a/modules/reitit-core/java-src/reitit/Trie.java#L181
Then Match would be immutable, which would prevent us from implementing more bugs like this in the future.

ikitommi:
Made the `Trie$Match` immutable.

PR #281, opened 2019-05-22
Fix a bunch of typos

malesch:


ikitommi:
Thanks! There was a lot of typos. Curious, did you have a tool to find all those?

malesch:
Not really. I stumbled over a typos while reading and then just went through the main files with IntelliJ. The editor then underlines unknown words but to be more efficient I would have to update the dictonary with some finnish expressions 😉 .

Issue #283, opened 2019-05-23
Create-default-handler expects all handlers or none


kardan:
## Issue
Configure an app with only not-found default handler like:
```
(create-default-handler {:not-found (constantly {:status 404, :body "", :headers {}}))
```
And then make a POST request to a non valid handler would throw an NPE
```
java.lang.NullPointerException:
    reitit.ring/create-default-handler/fn       ring.cljc:  160
              reitit.ring/ring-handler/fn       ring.cljc:  285
                                      ...
                      reitit.ring-test/fn  ring_test.cljc:  205
                 clojure.test/test-var/fn        test.clj:  716
                    clojure.test/test-var        test.clj:  716
            eftest.runner/test-vars/fn/fn      runner.clj:   90
             clojure.test/default-fixture        test.clj:  686
               eftest.runner/test-vars/fn      runner.clj:   85
    eftest.runner/wrap-test-with-timer/fn      runner.clj:   32
               eftest.runner/test-vars/fn      runner.clj:  104
             clojure.test/default-fixture        test.clj:  686
                  eftest.runner/test-vars      runner.clj:   99
                    eftest.runner/test-ns      runner.clj:  112
              eftest.runner/test-all/f/fn      runner.clj:  120
                      clojure.core/map/fn        core.clj: 2747
                                      ...
                        clojure.core/next        core.clj:   64
                     clojure.core/reduce1        core.clj:  936
                     clojure.core/reduce1        core.clj:  926
                  clojure.core/merge-with        core.clj: 3051
                  clojure.core/merge-with        core.clj: 3043
                                      ...
                       clojure.core/apply        core.clj:  659
                 eftest.runner/test-all/f      runner.clj:  121
                eftest.runner/test-all/fn      runner.clj:  123
              clojure.core/with-redefs-fn        core.clj: 7434
                   eftest.runner/test-all      runner.clj:  123
                  eftest.runner/run-tests      runner.clj:  185
 metosin.bat-test.impl/reload-and-test/fn        impl.clj:  161
metosin.bat-test.impl/maybe-run-cloverage        impl.clj:  125
    metosin.bat-test.impl/reload-and-test        impl.clj:  159
                metosin.bat-test.impl/run        impl.clj:  200
                            user/eval3355       REPL Input
                                      ...
                 clojure.main/load-script        main.clj:  278
                    clojure.main/init-opt        main.clj:  280
                  clojure.main/initialize        main.clj:  311
                    clojure.main/null-opt        main.clj:  345
[...]

PR #284, opened 2019-05-23
[#283] Allow to pass some defaults to create-default-handler

kardan:
To allow only to pass in a :not-found handler to the (ring/create-default-handler)


ikitommi:
Looks great, thanks!

PR #299, opened 2019-06-27
Correct doc (obsolete example of middleware/create)

rynkowsg:
`reitit.middleware/create `has been removed at [927d4d43](https://github.com/metosin/reitit/commit/927d4d43).
It seems `reitit.middleware/map->Middleware` is the right approach
instead.

More:
https://clojurians.slack.com/archives/C7YF1SBT3/p1561635497233300


ikitommi:
thanks!

PR #302, opened 2019-07-13
Make Map Destructuring of Namespaced Keys more Beautiful

alexanderkiel:
It's possible to put the :keys keyword in the namespace of the keys one likes to
destructure. With that one can use symbols in the vector again. One advantage of
having symbols is, that Cursive grays them out if not used. I found two
occurrences of unused destructured keys.


ikitommi:
Thanks!

PR #307, opened 2019-08-19
Update docs for controller and identity

jahson:
Fixed typos in a sentence about `start`. Also updated the information about `stop` call, looks like it will be called with last controller identity value, not return value of params


ikitommi:
Thanks!

PR #313, opened 2019-09-08
Fix event definitions for routes with parameters

dimovich:
The current implementation will fail for `(re-frame/dispatch [::navigate ::some-route {:some :params}])`.

Update the definitions of `::navigate` event and `::navigate!` effect to accept the full route arguments.


ikitommi:
Thanks!

Issue #205, opened 2019-01-16
`ring.middleware.session/wrap-session` doesn't work with reitit


ikitommi:
Reitit is a route-first architecture: the middleware/interceptor chain is created and optimized per endpoint and the chain is only applied if a route is matched.

The approach has a caveat: some middleware, like Ring's [wrap-session](https://github.com/ring-clojure/ring/blob/master/ring-core/src/ring/middleware/session.clj) don't work as expected: `wrap-session` instantiates the default memory store per instance. This means that **every route will have their own memory-store**. This is a surprising default and should be resolved. 

How to fix:

1) mount the `wrap-session` outside of the router so there is only one instance of the mw for the whole app. There is `:middleware` option in `ring-handler` for this:

```clj
(require '[reitit.ring :as ring])
(require '[ring.middleware.session :as session])

(defn handler [{session :session}]
  (let [counter (inc (:counter session 0))]
    {:status 200
     :body {:counter counter}
     :session {:counter counter}}))

(def app
  (ring/ring-handler
    (ring/router
      ["/api"
       ["/ping" handler]
       ["/pong" handler]])
    (ring/create-default-handler)
    ;; the middleware on ring-handler runs before routing
    {:middleware [session/wrap-session]}))
```

2) create a single session store and use it within the routing table (all instances of the session middleware will share the single store).

```clj
(require '[ring.middleware.session.memory :as memory])

;; single instance
(def store (memory/memory-store))

;; inside, with shared store
(def app
  (ring/ring-handler
    (ring/router
      ["/api"
       {:middleware [[session/wrap-session {:store store}]]}
       ["/ping" handler]
       ["/pong" handler]])))
```

3. Lift the `wrap-session` into Reitit Middleware & Interceptor and things just work.

```clj
(require '[reitit.ring.middleware.session :as session])

(def app
  (ring/ring-handler
    (ring/router
      ["/api"
       {:middleware [(session/session-middleware)]}
       ["/ping" handler]
       ["/pong" handler]])))
```

ikitommi:
Happy to take PRs for option 3.

ikitommi:
original: https://www.reddit.com/r/Clojure/comments/ad228o/sessions_under_the_leiningen_reagent_template/

valerauko:
2 doesn't work if the store has any mount state in it, it just won't get started

BinitaBharati:
Thanks, number 1 solution worked for me!

erkkikeranen:
[...]

Issue #264, opened 2019-05-02
Reitit 1.0.0


ikitommi:
This issue will contain some of the things that should be changed, but would be breaking changes. We could collect them for 1.0.0, which could out soon:

## Clear component lifecycle

All components (middleware, interceptors, option implementation) should be constructed separately, so there is clear separation of when things happen. Currently all interceptors work like this, e.g. `(coercion/coercion-interceptor)` but most middleware don't (`coercion/coercion-middleware`). This is kinda bad as there are shared components like `swagger/swagger-feature`, which one needs to read the source to see how it is defined. Currently, because of this, some things fail at runtime, and only at [triple-poop phase](https://www.slideshare.net/metosin/reitit-clojurenorth-2019-141438093/25), which is really bad.

Also, having a separate constructor allows specs to be attached to them, effectively validating all component creation at router creation time. This is good.

## Others

* consistent packaging
  * no plurals, e.g. `reitit.http.interceptors.*` => `reitit.http.interceptor.*`
* request & context diffing
  * rename `reitit.ring.middleware.dev/print-request-diffs` as it also diffs responses
  * should be in the `reitit-dev` module as they are not meant for production (deep diff + pretty printing is really slow).

## How to break

New package and/or ns? Just break it? We broke mostly everything with [compojure-api 1.0.0](https://www.metosin.fi/blog/compojure-api-100/), but there are also [thoughs on growing things](https://vlaaad.github.io/2019-04-05/growing-cljfx).

valerauko:
>New package and/or ns? Just break it?

I say just break it. Semantic versioning is there for a reason, and with a major version upgrade, reasonable breakage is to be expected

esuomi:
@valerauko Reitit doesn't use SemVer but Break Versioning instead... :) In this case that doesn't matter in practice though, but technically it could.

miikka:
A quick note: for the var renamings (`print-request-diffs`), we could do *right now* what the growing things post suggests: let's rename the var and add a `^:deprecated` alias for the old thing. If we want, we can later remove the deprecated aliases as a breaking change.

PR #316, opened 2019-10-01
Add ring session middleware

dawran6:
This PR addresses issue #205 

Problem: 
> wrap-session instantiates the default memory store per instance. This means that every route will have their own memory-store. This is a surprising default and should be resolved.

Solution: The `reitit.ring.middleware.session` ns shares one single store instance when the session store isn't specified from the user.

Please let me know if there's anything that I can improve for this PR. Thank you!

### DIFF
--- a/modules/reitit-middleware/src/reitit/ring/middleware/session.clj
+++ b/modules/reitit-middleware/src/reitit/ring/middleware/session.clj
@@ -1,10 +1,13 @@
 (ns reitit.ring.middleware.session
   (:require
+   [clojure.spec.alpha :as s]
    [ring.middleware.session :as session]
    [ring.middleware.session.memory :as memory]))
 
+(s/def ::spec (s/keys :opt-un [::session]))
+
 (def ^:private store
-  "The in-memory session store.
+  "The default shared in-memory session store.
 
   This is used when no `:session` key is provided to the middleware."
   (atom {}))
@@ -12,12 +15,19 @@
 (def session-middleware
   "Middleware for session.
 
-  Decodes the session from a request map into the `:session` value and updates the session store
-  based on the response's `:session` value.
+  Enter:
+  Add the `:session` key into the request map based on the `:cookies`
+  in the request map.
+
+  Exit:
+  When `:session` key presents in the response map, update the session
+  store with its value. Then remove `:session` from the response map.
 
   | key          | description |
   | -------------|-------------|
   | `:session`   | `ring.middleware.session.store/SessionStore` instance. Use `ring.middleware.session.memory/MemoryStore` by default."
   {:name    :session
-   :compile (fn [{:keys [session] :or {session {:store (memory/memory-store store)}}} _]
+   :spec    ::spec
+   :compile (fn [{:keys [session]
+                 :or   {session {:store (memory/memory-store store)}}} _]
               {:wrap #(session/wrap-session % session)})})
--- a/modules/reitit-middleware/src/reitit/ring/middleware/session.clj
+++ b/modules/reitit-middleware/src/reitit/ring/middleware/session.clj
@@ -1,0 +1,23 @@
+(ns reitit.ring.middleware.session
+  (:require
+   [ring.middleware.session :as session]
+   [ring.middleware.session.memory :as memory]))
+
+(def ^:private store
+  "The in-memory session store.
+
+  This is used when no `:session` key is provided to the middleware."
+  (atom {}))
+
+(def session-middleware
+  "Middleware for session.
+
+  Decodes the session from a request map into the `:session` value and updates the session store
+  based on the response's `:session` value.
+
+  | key          | description |
+  | -------------|-------------|
+  | `:session`   | `ring.middleware.session.store/SessionStore` instance. Use `ring.middleware.session.memory/MemoryStore` by default."
+  {:name    :session
+   :compile (fn [{:keys [session] :or {session {:store (memory/memory-store store)}}} _]
+              {:wrap #(session/wrap-session % session)})})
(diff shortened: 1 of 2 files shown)


modules/reitit-middleware/src/reitit/ring/middleware/session.clj
```diff
@@ -0,0 +1,33 @@
...
+  Add the `:session` key into the request map based on the `:cookies`
+  in the request map.
+
+  Exit:
+  When `:session` key presents in the response map, update the session
+  store with its value. Then remove `:session` from the response map.
+
+  | key          | description |
+  | -------------|-------------|
+  | `:session`   | `ring.middleware.session.store/SessionStore` instance. Use `ring.middleware.session.memory/MemoryStore` by default."
```
ikitommi: I think it would be good to use the same options as the underlaying mw: https://github.com/ring-clojure/ring/blob/master/ring-core/src/ring/middleware/session.clj#L85-L100

modules/reitit-middleware/src/reitit/ring/middleware/session.clj
```diff
@@ -0,0 +1,33 @@
+(ns reitit.ring.middleware.session
+  (:require
+   [clojure.spec.alpha :as s]
+   [ring.middleware.session :as session]
+   [ring.middleware.session.memory :as memory]))
+
+(s/def ::spec (s/keys :opt-un [::session]))
```
miikka: Didn't review this in detail, but the spec `::session` is not defined anywhere.

modules/reitit-middleware/src/reitit/ring/middleware/session.clj
dawran6: Thanks! I'm on it.

modules/reitit-middleware/src/reitit/ring/middleware/session.clj
dawran6: Thank you for noticing it. I actually didn't know how the `:spec` key is used for reitit middlewares and I was trying to follow along https://github.com/metosin/reitit/blob/master/modules/reitit-middleware/src/reitit/ring/middleware/muuntaja.clj#L6-L7.
In the next commit I'll indicate the `:session` is a map of https://github.com/ring-clojure/ring/blob/master/ring-core/src/ring/middleware/session.clj#L85-L100

modules/reitit-middleware/src/reitit/ring/middleware/session.clj
```diff
@@ -0,0 +1,38 @@
...
+  Add the `:session` key into the request map based on the `:cookies`
+  in the request map.
+
+  Exit:
+  When `:session` key presents in the response map, update the session
+  store with its value. Then remove `:session` from the response map.
+
+  | key          | description |
+  | -------------|-------------|
+  | `:session`   | A map of options that passes into the [`ring.middleware.session/wrap-session](http://ring-clojure.github.io/ring/ring.middleware.session.html#var-wrap-session) function`."
```
dawran6: @ikitommi I updated the docstring with a link to the `wrap-session` function. The spec of `:session` is also updated above.

dawran6:
I might have misunderstood the usage of `:spec` because I don't see any error when the `:session` route data is invalid. For example, the `:store` cannot be `nil`:

```clojure
(def app
  (ring/ring-handler
   (ring/router
    ["/api"
     {:session    {:store nil}
      :middleware [mw/session-middleware]}
     ["/ping" handler]
     ["/pong" handler]])))
```

dawran6:
> I might have misunderstood the usage of `:spec` because I don't see any error when the `:session` route data is invalid.

I forgot about the `:validate` key.

ikitommi:
Route data validation is not on by default, here's the guide: https://cljdoc.org/d/metosin/reitit/0.3.9/doc/ring/route-data-validation

dawran6:
Hi @ikitommi @miikka, thanks for helping. I did another force push to fix another spec issue and include the test for the spec validation. Would you review it again when you have the time? Thanks.

modules/reitit-middleware/.nrepl-port
```diff
@@ -0,0 +1 @@
+43803
```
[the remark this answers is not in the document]
dawran6: Of course. I'm on it!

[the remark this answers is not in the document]
dawran6: Done. Thanks again to catch this.

ikitommi:
One thing that comes to mind is that how should the middleware be configured. In reitit, there are options:

1. **via middleware options**
pass the store etc. options for a `create` function, that is used to create the middleware. Interceptors use this is a lot and I think reitit 1.0.0 should do this too: it adds a clear phase to initialize one middleware **once** per application. Not the best option per today.

2. **via route-data, on by default**
if mw is mounted, it's on. Mounting a session-mw into top-level means there is no way to disable it for a certain route. Not optimal.

3. **via route-data, off by default**
mounting mw add just capabilities for a route. Only after adding a `:session` route data, enables it for certain routes. Most reitit-internal mw work like this, for example, content-negotiation & coercion.

would be good to have some coherence here. Just not sure what is the best way to do this. Comments?

dawran6:
I agree that **option 2** is not optimal, so this PR will to be fixed.

My personal preference is **option 3**. I value the consistancy with other internal mw (**option 3**) more than the clarity from **option 1**.

dawran6:
The commit a3c64baeba60f9330da001c72b579b5425b7c9bf changes from option 2 to option 3. I think we need some more feedbacks to make a decision.

I think another reason to support option 3 (in general) is that the router code would be cleaner if anyone wants to have different configurations on routes. Say:

```clojure
["/routes"
  {:middleware [session/session-middleware]}
 ["/with-session-mem!"  {:session {:store mem-store}} ...]
 ["/with-session-cookie!"  {:session {:store cookie-store}} ...]
 ["/without-session!" ...]]
```

Versus

```clojure
["/routes"
 ["/with-session-mem!"  {:middleware [(session/session-middleware {:store mem-store})]} ...]
 ["/with-session-cookie!"  {:middleware [(session/session-middleware {:store cookie-store})]} ...]
 ["/without-session!" ...]]
```

dawran6:
I finally see this today: #264. Perhaps put this PR on hold till #264 makes a decision.

Issue #265, opened 2019-05-08
Example project with integrant


ikitommi:
integrant + reitit-ring.

gdrte:
http://Luminusweb.net generates reitit-ring template by default. Do you want anything different?

ikitommi:
Luminus should be mentioned in the (non-existent) "getting started" guide and it surely could be used to create a base (full-stack?) for the sample project under `/examples`. Would you like to do a PR?

For the future: there is a lot of roof for deeper integration with Integrant, e.g. apply the closed spec validation & use the reitit error messages with integrant (it has hooks for spec-validation already).

gdrte:
I can do an example, but I may come for your help often.

dawran6:
> For the future: there is a lot of roof for deeper integration with Integrant, e.g. apply the closed spec validation & use the reitit error messages with integrant (it has hooks for spec-validation already).

@ikitommi Would you elaborate a bit more about this? What I think of, yet never tried, is to put the route data into integrant keys. Then I guess I can spec-validate the keys with `reitit.ring.spec/data` using `integrant.core/pre-init-spec` like [this](https://github.com/weavejester/integrant#specs).

ikitommi:
I'm not a fan of spreading the partial route definitions under integrant, too much IoC happening there. We are using Integrant only for top-level stateful things all reitit stuff is under the one http-handler component.

So, just a small example of integrant (with just http-handler for example) + reitit-ring would be nice to add to examples.

The grand plan (outside of this issue): use [virhe](https://github.com/metosin/virhe) for both Integrant + Reitit error messages so that there is a coherent developer experience across the whole stack. Sadly, spec can't be used here yet: nested specs can't be closed with spec-tools with Spec1.

Currently we are working with [malli](https://github.com/metosin/malli), that will solve all the issues with closing & merging nested specs and most likely will port all the reitit internal specs into schemas as package the coherent DX into [talvi](https://github.com/metosin/talvi). If Spec2 comes out and does what is needed, can use that too. We'll see.

dawran6:
@ikitommi would you give me some feedbacks on https://github.com/dawran6/reitit/tree/integrant-example/examples/ring-integrant before I submit the PR? Thank you.

ikitommi:
Done, big thanks!

PR #322, opened 2019-10-17
Add an example project with Integrant

dawran6:
This PR addresses #265 

Feedbacks are welcome.
Currently, the project is minimal. The system only contains two keys: `:example/jetty` and `:example/handler`. I can certainly add more features.

### DIFF
--- a/examples/ring-integrant/.gitignore
+++ b/examples/ring-integrant/.gitignore
@@ -1,0 +1,11 @@
+/target
+/classes
+/checkouts
+pom.xml
+pom.xml.asc
+*.jar
+*.class
+/.lein-*
+/.nrepl-port
+.hgignore
+.hg/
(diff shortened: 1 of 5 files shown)


ikitommi:
Hi. Looks great! What do you think of adding [integrant repl](https://github.com/weavejester/integrant-repl) to the mix. We use that, makes the developer Experience much better, when there are more namespaces to cover.

dawran6:
Yup! I like the idea. I use Integrant repl myself. It makes sense to also show the whole developer experience in the example.

dawran6:
The `integrant/repl` related functions are moved to the `user` ns.
I also removed the `example.server/server` atom since `integrant/repl` tracks the system state for me. Thanks for reviewing @ikitommi !

ikitommi:
Awesome, thanks!

PR #323, opened 2019-10-21
Update compiling_middleware.md regarding specs

ccann:
Hello,

Please consider this documentation PR inspired by some confusion I encountered during my first foray into writing middleware for Reitit. Hopefully I've used the correct language and the docs are accurate. :-)

### DIFF
--- a/doc/ring/compiling_middleware.md
+++ b/doc/ring/compiling_middleware.md
@@ -70,3 +70,43 @@ To demonstrate the two approaches, below are response coercion middleware writte
 ```
 
 It has 50% less code, it's much easier to reason about and is much faster.
+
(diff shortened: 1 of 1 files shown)


ikitommi:
Thanks for the docs! If you want to enforce the specs despite the fact that the middleware is not doing anything at request processing time, you should return a map without a `:wrap` or `:compile` keys, e.g. an empty map. This way, the mw (and the specs) are still in the chain, just not part of request processing. Hidden docs here: https://github.com/metosin/reitit/blob/master/modules/reitit-ring/src/reitit/ring/coercion.cljc#L21-L43

could you add that to the docs?

ccann:
@ikitommi Added, let me know what you think :-)

Issue #175, opened 2018-11-17
Allow route expansion to be configured to publish intermediate paths


ikitommi:
Currently, only leaf routes are used as endpoints. Intermediate paths require and extra empty path fragment, so we have to write:

```clj
["/users"
 {:interceptors [(require-role :admin)]}
 
 [""
  {:get {:handler list-users}}]

 ["/:id"
  {:get {:handler get-user}
   :delete {:handler delete-user}}]]
```

Router could be configured with a new option like `:endpoint` which would take the path + route data and return truthy if that  path can be used as endpoint. Ring-router could have default `:handler` allowing the previous example to be written as:

```clj
["/users"
 {:interceptors [(require-role :admin)]
  :get {:handler list-users}}

 ["/:id"
  {:get {:handler get-user}
   :delete {:handler delete-user}}]]
```

Comments welcome.

tonivanhala:
One edge case (?) that I noticed while working on this: currently, leaf routes may inherit their ancestor’s handlers, as they are merged from above.

I’m thinking of preserving the inheritance of handlers in the default endpoint validation, so that: 1) leaf nodes always act as valid routes, if they either a) have their own handler, or b) inherit a handler from ancestor, and 2) intermediate paths are valid routes, iff they have their own valid handler. So, intermediate paths would still not be valid, if they only inherit a handler.

Not sure if the logic is easy to grasp. Maybe I should write docs for this as well.

ikitommi:
leaf-nodes don't have to have a handler, direct or inherited: the response can be returned from a middleware. There should be a test to see that both mw & int chains contain at least 1 thing, could be a handler.

ikitommi:
Related commit: https://github.com/metosin/reitit/commit/c3de6ff3ddc6a62d139a31a089ab1423876bbdc0

tonivanhala:
Re: leaf nodes. Oh right, I did not realize that. 

I think checking that at least one thing is in leaf route should be another issue. More than 50 tests seem to depend on current functionality, i.e., leaves always being valid routes. 

I'm drafting a PR where 1) leaves are routes and 2) intermediate paths are checked for either `:handler` (backend), or `:view` or `:name` (frontend). This check is made at `resolve-routes` stage, before checking for conflicting routes etc. Currently, intermediate routes do not pass that stage. 

More in the PR.

Issue #271, opened 2019-05-14
Reverse-routing with ring/http & request methods


ikitommi:
Reverse routing uses the `:name` route data, which is a concept of the core router.  Currently it doesn't understand the ring method leaves like `:get`, `:post` etc. This mean that there is no way to identify each unique path+method pair. Suggestions:

# customize reverse routing

- make a router options that allows route `:name`s to be collected & reverse routed via custom functions
- modify `reitit.ring/router` & `reitit.http/router` set the value the reverse routing works from both leaves & top-level
- BAD: resulting Match of a reverse routing should somehow have both the path & relevant route data parts

# higher level reverse routing with ring/http

- Swagger defines `operationId` per method endpoint. We could have just `:id`. 
- New `HttpRouter` protocol with `match-by-id` & `route-id`s for both ring & http routers to point to method endpoints
- `match-by-id` would return the method as a new key in `Match`

RokLenarcic:
I'd say that second solution looks more neatly packaged into existing concepts. 

The first suggestion is more general with the collecting functions, but it's not clear what other uses this would address besides HTTP methods. And it's hard to design a general facility without having multiple usecases of the bat.

rschmukler:
I'd like to propose that `match-by-name` just be extended to behave as `match-by-id` instead. Introducing `match-by-id` introduces more surface area and potential for some confusing behaviors. What happens if a route has an `:id` and a `:name`? Why enforce `:name` be unique when we have `:id`? Maybe I am missing something, but I'd ask in which cases would `:name` and `:id` not behave the same, and, is the additional complexity worth having those two distinct behaviors?

For my specific use case I wanted to share the routing table from my back-end with my front-end to make calls to the API + handle resolution of specs for request and response formats. Eg imagine something like:

```clj
(def routes
  (http/router
    ["/api"
      ["/users"
        [["" {:post {:name       :api.users/create 
                     :parameters {:body :user/new}
                     :responses   {201 {:body {:data {:key string?}}}}]]]))
```

I'd like to be able to define a function like:

```clj
(api/make-request :api.users/create {:username "Bob" :password "Foo"})
```

Which can automatically figure out the request / response specs (for `st/coerce` + the HTTP method + path.

ikitommi:
For 1, some extra considerations:

1) how would the resulting `Match` indicate what `:request-method` was matched?
   a) new key, `:request-method` which can be `nil` (top-level match) or a keyword (method-matched)
[...]

PR #326, opened 2019-10-25
Publish intermediate paths by default

tonivanhala:
Currently, only leaf routes can be used as endpoints. This causes unexpected behaviour, for example, when intermediate paths are defined with a handler, but they are not published. This seems to be confusing, especially when starting out with Reitit. 

This PR changes the default so that intermediate paths are published as endpoints, if they have certain data (e.g., a `:handler` for a backend route) that is typical for intended endpoints. 

Changes in this PR allow customizing the endpoint resolution with a `:endpoint?` predicate, and provide default predicates for the frontend and Ring routing. 

See https://github.com/metosin/reitit/issues/175

Note that checking leaf routes for any data is left for later implementation.

### DIFF
--- a/modules/reitit-core/src/reitit/core.cljc
+++ b/modules/reitit-core/src/reitit/core.cljc
@@ -334,7 +334,6 @@
 ;; Creating Routers
 ;;
 
-
 (defn ^:no-doc default-router-options []
   {:lookup (fn lookup [[_ {:keys [name]}] _] (if name #{name}))
    :expand expand
--- a/modules/reitit-core/src/reitit/core.cljc
+++ b/modules/reitit-core/src/reitit/core.cljc
@@ -334,13 +334,15 @@
 ;; Creating Routers
 ;;
 
+
 (defn ^:no-doc default-router-options []
   {:lookup (fn lookup [[_ {:keys [name]}] _] (if name #{name}))
    :expand expand
    :coerce (fn coerce [route _] route)
    :compile (fn compile [[_ {:keys [handler]}] _] handler)
    :exception exception/exception
-   :conflicts (fn throw! [conflicts] (exception/fail! :path-conflicts conflicts))})
+   :conflicts (fn throw! [conflicts] (exception/fail! :path-conflicts conflicts))
+   :endpoint  impl/leaf-endpoint})
 
 (defn router
   "Create a [[Router]] from raw route data and optionally an options map.
@@ -360,7 +362,10 @@
   | `:validate`  | Function of `routes opts => ()` to validate route (data) via side-effects
   | `:conflicts` | Function of `{route #{route}} => ()` to handle conflicting routes
   | `:exception` | Function of `Exception => Exception ` to handle creation time exceptions (default `reitit.exception/exception`)
-  | `:router`    | Function of `routes opts => router` to override the actual router implementation"
+  | `:router`    | Function of `routes opts => router` to override the actual router implementation
+  | `:endpoint`  | Function of `prev-path path expanded-data data children => {:endpoint :inherit}`.
+  |              | Used to transform data for this endpoint and data inherited to child routes.
+  |              | Return falsy or `{:endpoint nil}` to skip path as endpoint and pass data to children."
   ([raw-routes]
    (router raw-routes {}))
   ([raw-routes opts]
--- a/modules/reitit-frontend/src/reitit/frontend.cljs
+++ b/modules/reitit-frontend/src/reitit/frontend.cljs
@@ -2,7 +2,8 @@
   (:require [clojure.set :as set]
             [reitit.coercion :as coercion]
             [reitit.coercion :as rc]
-            [reitit.core :as r])
+            [reitit.core :as r]
+            [reitit.impl :as impl])
   (:import goog.Uri
            goog.Uri.QueryData))
 
@@ -45,13 +46,17 @@
   ([router name path-params]
    (r/match-by-name router name path-params)))
 
+(def frontend-endpoint (impl/mk-intermediate-endpoint-transform [{:kss       [[:name]]
+                                                                  :transform :consume}]))
+
 (defn router
   "Create a `reitit.core.router` from raw route data and optionally an options map.
   Enables request coercion. See [[reitit.core/router]] for details on options."
   ([raw-routes]
    (router raw-routes {}))
   ([raw-routes opts]
-   (r/router raw-routes (merge {:compile rc/compile-request-coercers} opts))))
+   (r/router raw-routes (merge {:compile rc/compile-request-coercers
+                                :endpoint frontend-endpoint} opts))))
 
 (defn match-by-name!
   "Logs problems using console.warn"
--- a/modules/reitit-ring/src/reitit/ring.cljc
+++ b/modules/reitit-ring/src/reitit/ring.cljc
@@ -65,6 +65,11 @@
       ([request respond _]
        (respond (handle request))))))
 
+(def ring-endpoint (impl/mk-intermediate-endpoint-transform [{:kss (for [method http-methods] [method :handler])
+                                                              :transform :inherit}
+                                                             {:kss       [[:name]]
+                                                              :transform :consume}]))
+
 ;;
 ;; public api
 ;;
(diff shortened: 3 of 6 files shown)


modules/reitit-core/src/reitit/core.cljc
```diff
@@ -334,13 +334,36 @@
...
 ;;
 
+(defn leaf-endpoint?
+  [_ _ _ _ childs]
+  (not (seq childs)))
+
+(defn mk-intermediate-endpoint-predicate
+  "Create predicate that returns truthy for valid endpoints, where the path is a leaf in
+  the route tree, or the route data contains one of the specified keys at top-level or
+  at the first level of nested maps.
```
ikitommi: Looking at nested maps can muddle with user specific data, e.g.

```clj
["/topic" {:topic {:name "iot.inbox"}}]
```

modules/reitit-ring/src/reitit/ring.cljc
```diff
@@ -65,6 +65,8 @@
       ([request respond _]
        (respond (handle request))))))
 
+(def ring-endpoint? (r/mk-intermediate-endpoint-predicate #{:handler}))
```
ikitommi: would `:name` be an endpoint? routes don't have to have a handler...

modules/reitit-frontend/src/reitit/frontend.cljs
```diff
@@ -38,13 +38,16 @@
   ([router name path-params]
    (r/match-by-name router name path-params)))
 
+(def frontend-endpoint? (r/mk-intermediate-endpoint-predicate #{:name :view}))
```
ikitommi: is `:view` used by reitit?

ikitommi:
what if there is a endpoint key, e.g. `:name` on a intermediate route, which has multiple intermediate/leaf routes, do the values inherit into all of those too? Might be confusing if a same `:name` is found in multiple route data. 

- Should `:name` be "consumed" by the intermediate route so it's not present in the child route data?

- Should the handler be inherited to child routes as there is no unique constraint for it

- if there would be two ways to handle the intermediate endpoints, how should it be configured? Maybe with map: `:endpoints {:name :consume, :handler :inherit}` and a separate key :reitit.core/endpoint-lookup` which would default to `indentity` (from top-level route data) and ring would override this to do lookup also under method maps.

modules/reitit-frontend/src/reitit/frontend.cljs
tonivanhala: No. Removed it from defaults.

modules/reitit-ring/src/reitit/ring.cljc
tonivanhala: Changed it so that `:name` signifies endpoint and is consumed by the intermediate path, i.e., not passed on to child routes.

modules/reitit-core/src/reitit/core.cljc
tonivanhala: True. Changed it so that the default is to check for handler defined under http methods.

tonivanhala:
Changed the endpoint resolution so that the custom endpoint function is a transformer instead of a predicate. Data for the endpoint and data passed to children can be transformed, while resolving intermediate routes. Default transformation is to consume `:name` and pass `:handler` to children.

modules/reitit-ring/src/reitit/ring.cljc
```diff
@@ -65,6 +65,11 @@
       ([request respond _]
        (respond (handle request))))))
 
+(def ring-endpoint (r/mk-intermediate-endpoint-transform [{:kss (for [method http-methods] [method :handler])
+                                                           :transform :inherit}
+                                                          {:kss       [[:name]]
+                                                           :transform :consume}]))
+
```
ikitommi: Contract for `mk-intermediate-endpoint-transform` is not defined and looks bit complex. Ring already has a custom `:compile` that knows how to handle the http-methods. Would be great if that could be used instead. Does the `:compile` happen too late so it can't be used here? If it could, all the routers would work the same way and we could just declare the `:endpoint` options as data, instead of a function callback.

ikitommi: also, there are comments that the `:name` [should be allowed only on leafs](https://github.com/metosin/reitit/issues/271#issuecomment-546647754). If that would be implemented, it would make this much simpler?

test/cljc/reitit/ring_test.cljc
```diff
@@ -116,7 +116,43 @@
...
+                    [""       {:name ::other-root}] ;; Conflicts with ::root path
+                    ["foo"    {:name ::foo}]
+                    ["bar"    {:get {:handler handler} :name ::bar}
+                     ["/baz"  {:get {:handler handler} :name ::baz}]]
+                    ["ping"   {:name ::ping}
+                     ["/pong" {:get {:handler handler} :name ::pong}]]]]
+        (is (thrown-with-msg?
+              ExceptionInfo
+              #"Router contains conflicting route paths"
+              (ring/router routes)))))))
```
ikitommi: great tests!!

modules/reitit-ring/src/reitit/ring.cljc
tonivanhala: > Contract for `mk-intermediate-endpoint-transform` is not documented and looks bit complex. 

`mk-intermediate-endpoint-transform` has a docstring which tells how it works. In any case, it is an implementation detail, which should be out of user's eyes. I agree that it seems complex, but I was struggling how to express (tersely) what kind of transform will be performed on different key seqs of route data. For reference, this implements the idea, that @ikitommi suggested, that some route data would be consumed by the intermediate route, while other would be passed down to child routes. 

The actual implementation of the endpoint transformation (which `mk-intermediate-endpoint-transform` creates for the new default endpoint resolution) is documented here: https://github.com/metosin/reitit/pull/326/files#diff-7d84e60643070c30a98ce39e202d308bR426

> reitit-ring already has a custom `:compile` that knows how to handle the http-methods. Would be great if that could be used instead. Does the `:compile` happen too late so it can't be used here? 

Currently, the intermediate routes do not survive the first step of endpoint (route) resolution, so the compile is too late with current behaviour. I'm pretty sure I wrote that somewhere, but I'm not able to find it. 

> If it could, all the routers would work the same way and we could just declare the `:endpoint` options as data, instead of a function callback.

Hmm. This is kind of what I was aiming for... Would it be better to have the "endpoint transform" as completely internal to route resolution, and have router-specific configuration as just data: ?
```
(def ring-endpoint [{:kss (for [method http-methods] [method :handler])
                               :transform :inherit}
                    {:kss       [[:name]]
                     :transform :consume}]))
```

tonivanhala: > also, there are comments that the :name should be allowed only on leafs. If that would be implemented, it would make this much simpler?

I can't see where this is mentioned in the conversation? 

This (allowing `:name` only on leaves) would defeat the purpose of this whole PR, which is to enable intermediate routes. The rule that is applied in this PR is that paths with `:name` will likely be referenced by the user, so they can act as routes (and matched by name). 

However, the behaviour implemented in this PR (that you @ikitommi suggested) does seem to reduce conflicts with `:name`, as the default behaviour would now be to consume the name at the path. Currently, it is passed it to all children, which leads to duplicate names for routes. 

Maybe we could talk this through, so we're on the same map on what should be accomplished here.

test/cljc/reitit/ring_test.cljc
tonivanhala: Thanks! I hope they are, since this PR would change default route resolution... :grimacing:

modules/reitit-ring/src/reitit/ring.cljc
tonivanhala: Also, I would be more than happy for simpler, but concise, way of defining which paths are routes (predicate "endpoint?") while specifying the transform of route data (`consume` vs. `inherit`). :sweat_smile:

rschmukler: @tonivanhala I believe @ikitommi is referencing https://github.com/metosin/reitit/issues/271 when mentioning the disucssion around `:name`. I do agree that the issues discussed in there and this PR are somewhat contradictory.

One potential way to bring them back in sync might be to assume a `:method :get` configuration for an intermediate that has a `:name` and doesn't specify the `:method` itself. I believe that if that were the case, all of the changes mentioned in #271 could still be implemented on top of this PR.

I have run into the confusing behavior described and resolved by this PR so I definitely see value in it. @ikitommi thoughts on if the behavior in this PR is desired given your comments in #271 ?

tonivanhala:
Rebased this to keep up with current master, and cleaning up the history. 
Moved the new endpoint functions from `core` namespace to `impl`

tonivanhala:
Closing my PR which has been stale for five years.

PR #328, opened 2019-10-29
Workaround for CircleCI problems

miikka:
For some reason the CircleCI config from #327 do not work in master. I wonder if it's somehow related to the fact that `build-docs` is enabled for master but not for other branches. See e.g. [this workflow run](https://circleci.com/workflow-run/67522e3a-8ae6-4791-821a-88f804b1d745) for commit 54dc6e3dabc10a264c0a7d2a62417fe8d4ff7f6d.

I couldn't find anything from Circle's docs. As a workaround, I preprocessed the config manually with the `circleci` command-line tool:

```shell
cp .circleci/config.yml .circleci/config.orig.yml
circleci config process .circleci/config.orig.yml > .circleci/config.yml
```

cc. @valerauko

### DIFF
--- a/.circleci/config.orig.yml
+++ b/.circleci/config.orig.yml
@@ -1,0 +1,130 @@
+version: 2.1
+
+jobs:
+  test-clj:
+    working_directory: ~/test
+    parameters:
+      image-name:
+        type: string
+    docker:
+      - image: << parameters.image-name >>
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - 'v1-clj-{{ checksum "project.clj" }}'
+            - 'v1-clj-'
+            - 'v1-test-'
+      - run:
+          name: Install modules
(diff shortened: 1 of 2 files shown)


valerauko:
This is very weird. Why would the same config work for other branches but not on master? It's a config failure even...

valerauko:
So it seems that the jobs of the PR in question were run on my account, and not on the one controlling the reitit repo. https://circleci.com/gh/valerauko/reitit/tree/test-multiple-jdk

Could this issue be related to this "Enable pipelines" config setting? https://circleci.com/docs/2.0/reusing-config/#getting-started-with-config-reuse

miikka:
> Could this issue be related to this "Enable pipelines" config setting? https://circleci.com/docs/2.0/reusing-config/#getting-started-with-config-reuse

That's it! Thanks! I skipped this section and forgot that they have to be enabled per-project basis.

PR #329, opened 2019-11-04
Manually tag conflicting routes

### LINKED ISSUE #324: Manually tagging conflicting routes
Currently, if there are few conflicting routes, one has to disable the conflict resolution to make it work. This does not effect performance of non-conflicting routes, but drops an important guard for the future (of accidentally making route conflicts). We could have a route data marker, e.g. `:conflicting` that could be used to mark that route is known to conflict. The conflict resolution could take this into account and only fail on non-tagged conflicts.

ilmoraunio:
Fixes #324 

Introduces a new route data key `:conflicting` which takes a logically true value and, upon true, excludes the route from the path conflicts report. The report is used during router initialization in conjunction with the `:conflicts` check to see if there are still any paths that are not ignored and are conflicting.

### DIFF
--- a/doc/basics/route_conflicts.md
+++ b/doc/basics/route_conflicts.md
@@ -2,7 +2,7 @@
 
 We should fail fast if a router contains conflicting paths or route names. 
 
-When a `Router` is created via `reitit.core/router`, both path and route name conflicts are checked automatically. By default, in case of conflict, an `ex-info` is thrown with a descriptive message. In some (legacy api) cases, path conflicts should be allowed and one can override the path conflict resolution via `:conflicts` router option.
+When a `Router` is created via `reitit.core/router`, both path and route name conflicts are checked automatically. By default, in case of conflict, an `ex-info` is thrown with a descriptive message. In some (legacy api) cases, path conflicts should be allowed and one can override the path conflict resolution via `:conflicts` router option or via `:conflicting` route data.
 
 ## Path Conflicts
 
@@ -25,14 +25,14 @@ Creating router with defaults:
 (r/router routes)
 ; CompilerException clojure.lang.ExceptionInfo: Router contains conflicting route paths:
 ;
-;    /:user-id/orders
+; -> /:user-id/orders
 ; -> /public/*path
 ; -> /bulk/:bulk-id
 ;
-;    /bulk/:bulk-id
+; -> /bulk/:bulk-id
 ; -> /:version/status
 ;
-;    /public/*path
+; -> /public/*path
 ; -> /:version/status
 ;
 ```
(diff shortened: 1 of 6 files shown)


doc/basics/route_conflicts.md
```diff
@@ -2,7 +2,7 @@
 
 We should fail fast if a router contains conflicting paths or route names. 
 
-When a `Router` is created via `reitit.core/router`, both path and route name conflicts are checked automatically. By default, in case of conflict, an `ex-info` is thrown with a descriptive message. In some (legacy api) cases, path conflicts should be allowed and one can override the path conflict resolution via `:conflicts` router option.
+When a `Router` is created via `reitit.core/router`, both path and route name conflicts are checked automatically. By default, in case of conflict, an `ex-info` is thrown with a descriptive message. In some (legacy api) cases, path conflicts should be allowed and one can override the path conflict resolution via `:conflicts` router option or via `:conflicting?` route option.
```
miikka: For consistency, it's better to call it route data. Route option is super-easy to mix up with router option.

doc/basics/route_conflicts.md
```diff
@@ -65,7 +67,21 @@ To just log the conflicts:
 ; /public/*path
 ; -> /:version/status
 ;
-; => #object[reitit.core$linear_router$reify]
+; => #object[reitit.core$quarantine_router$reify]
+```
+
+Alternatively, you can ignore conflicting paths individually via the `:conflicting?` option:
```
miikka: Ditto about route data.

ikitommi:
My first impression on this was that all conflicting routes should be tagged, e.g. both of: `[["/a]  ["/a"]]`. If this this would be easier for the end user to understand (and later: tooling to pick up) and it might make the impl code bit simpler. What do you think?

ikitommi:
could the key be `:conflicting`? There are few router option keys that end with `?`, but should be without (will brake those for 1.0.0) as `?` indicates a predicate function, not a Boolean value.

ikitommi:
Otherwise, looking really good

ilmoraunio:
> My first impression on this was that all conflicting routes should be tagged, e.g. both of: [["/a] ["/a"]]. If this this would be easier for the end user to understand (and later: tooling to pick up) and it might make the impl code bit simpler. What do you think?

Do you mean if I tag `[["/:a" {:conflicting true}] ["/b"]]`, then the `:conflicting` marker is assumed to spread to all conflicts of `/:a`?

Why I went with the current implementation was that I saw how exceptions are reported, namely:

```
(def routes
  [["/:a"] ["/b"]])
=> #'user/routes
(r/router routes)
Execution error (ExceptionInfo) at reitit.exception/exception (exception.cljc:19).
Router contains conflicting route paths:

   /:a
-> /b
```

I took that the user is supposed to act upon the problematic routes – or at least it would be odd if addressing the route's problem directly would not be possible. However, I don't see it problematic to have a bi-directionality aspect to `:conflicting`, ie. that both `[["/:a" {:conflicting true}] ["/b"]]` and `[["/:a"] ["/b" {:conflicting true}]]` don't throw. (If that is what you meant in the first place.)

> could the key be :conflicting?

I'll go with `:conflicting`. I'd like to avoid force-pushing because of this if possible. I wonder if commit messages are editable at squash time?

ilmoraunio:
Okay, re-reading with thought, you probably meant that `[["/a"] ["/a"]]` should be marked as `[["/a" {:conflicting true}] ["/a" {:conflicting true}]]`. That's possible, but then the exception reporting should probably be adjusted to show that both routes should be marked as conflicting, eg:

```
Router contains conflicting route paths:

-> /:a
-> /b
```

For the record, I did initially scope out on whether we should instruct the user during exception reporting on how to address conflicting paths. Having helpful exception messages is one possible strategy of helping the end user here. For example, the message could be:

```
Router contains conflicting route paths:

-> /:a
-> /b

To resolve the conflicts, either mark conflicting data routes with {:conflicting true} or override :conflicts for router.
```

We can also discuss this over at #324.

ilmoraunio:
The new version is available for review based on the discussion over at #324. Unfortunately, had to resort to force-pushing after all.

ilmoraunio:
Not sure this is something we want:

```clj
(r/router [["/:a"
            ["/a"]
            ["/a"]]])
Execution error (ExceptionInfo) at reitit.exception/exception (exception.cljc:19).
Router contains conflicting route paths:

-> /:a/a 
-> /:a/a 

(r/router [["/:a" {:conflicting true}
            ["/a"]
            ["/a"]]])
=> #object[reitit.core$quarantine_router$…]
```

ilmoraunio:
Alright, so above happens due to the way route data is passed to child nodes. I'll add a couple of tests to demonstrate.

modules/reitit-core/src/reitit/impl.cljc
```diff
@@ -85,6 +85,15 @@
             routes)
       (not-empty)))
 
+(defn unresolved-conflicts [path-conflicting]
+  (-> (into {}
+            (remove (fn [[[_ route-data] conflicts]]
+                      (and (:conflicting route-data)
+                           (every? (comp :conflicting second)
+                                   conflicts)))
+                    path-conflicting))
```
miikka: [golf] You could use a transducer here for tiny performance gains. `(into {} (remove ...) path-conflicting)`.

ilmoraunio: Done.

PR #332, opened 2019-11-15
Frontend: Multi-valued query params

nilern:
Make `"/foo?mode=foo&mode=bar"` produce `:query-params {:mode ["foo", "bar"]}` instead of
`:query-params {:mode "foo"}` by using `goog.Uri.QueryData#getValues` instead of `#get`.


ikitommi:
Looks good, feel free to merge/rebase. Could you update docs (if needed) & CHANGELOG too?

nilern:
I realized that round-tripping to query params and back will not work if we merge this since I did not touch the corresponding encoding code. Should probably do that.

nilern:
Implemented encoding in core. Seems that decoding already works on server side since query parameters are handled with Ring middleware there.

ikitommi:
parameter decoding via ring mw is not optimal in perf, or features, but a separate issue

ikitommi:
Looks great!!

PR #341, opened 2019-12-02
Implement malli-based coercion

ikitommi:
* Adds support for [malli-based](https://github.com/metosin/malli) coercion
* Strips extra keys on all EDN & Transit request bodies with clojure.spec (a MINOR bump?)

## TODO

* [x] encode on way back
* [x] strip-extra-keys on both ways?
* [x] example
* [x] docs

## Current defaults:

```clj
; defaults (closed + strip extras)
(reitit.malli.coercion/create)

;; open schemas, strip extra keys
(reitit.malli.coercion/create {:compile identity})

;; open schemas, don't strip extra keys
(reitit.malli.coercion/create {:compile identity, :strip-extra-keys false})
```

comments welcome on options & defaults


rschmukler:
+1 for striping extra keys both ways

rschmukler:
Just testing this out on my own project. A few things of note:

1) Right now it's impossible to pass options to malli directly, so for example, a custom registry can't be done. I recommend adding `malli-opts` to the root options.

2) Coercion and validation needs to happen in inverse order on response encoding. ie. We want to first ensure that the response is valid, and then transform it - right now we transform and the validate (which makes sense for decode).

ikitommi:
1) added `:options` key to coercion options

2) changed the order

3) short-cut syntaxes are root of many evil, but adding a extension fn to support this would add the evil to user-space. Converting a spec app to malli-based: https://github.com/metosin/reitit/pull/341/commits/3b53efc4398e68af9bfe387c66950fb5fec7c506

Added also an option `:error-keys` to control what kind of errors are created, oob support for `:humanized`:

Screenshot 2019-12-28 at 9 39 38

ikitommi:
stripping extra keys on return by default for all schemas... not sure if that is a good property.

ikitommi:
Thanks for the comments.

* humanized errors will be changed in: https://github.com/metosin/malli/pull/155

* The stripping-extra-keys on responses... not happy either way as changing a transformer introduces a lot of boilerplate, this should be resolved before releasing this. Possibilities:

### 1) By default, only strip extra keys from `:closed` schemas

* Schemas will have [`:closed`](https://github.com/metosin/malli/issues/31) property, effecting validation.
* `coercion` could have an option `:close-schemas` to close all (api) schemas. This can easily be set [programmatically](https://github.com/metosin/malli/issues/63) with malli.

### 2) Helper option in `coercion` 

* `:strip-extra-keys` to easily set the stripping on/off (without needing to re-create all the transformers)

### 3) Butt-naked

* The current way, one needs to re-wire the transformers to enable & disable stripping.

rschmukler:
My personal taste would be to lean towards #1. Making closed schemas a first-class citizen in malli seems most useful as it allows the validators / encoders / decoders to be used at the serialization boundaries of an application, which would be great. Similarly, adding the option (`:close-schemas`) also feels quite useful for ad-hoc generation of closed variants of schemas.

We could also have a function `(m.utils/closed schema)` which returns a new version of a schema with `{:closed true}` set in the properties.

ikitommi:
Let's go with 1, related PR on malli: https://github.com/metosin/malli/pull/156

ikitommi:
Default options are now:

```clj
(def default-options
  {:transformers {:body {:default default-transformer-provider
                         :formats {"application/json" json-transformer-provider}}
                  :string {:default string-transformer-provider}
                  :response {:default default-transformer-provider}}
   ;; set of keys to include in error messages
   :error-keys #{:type :coercion :in :schema :value :errors :humanized #_:transformed}
   ;; schema identity function
   :compile mu/closed-schema
   ;; strip-extra-keys (effects only default transformers!)
   :strip-extra-keys true
   ;; add default values
   :default-values true
   ;; malli options
   :options nil})
```

we have an internal review of malli on friday, small tuning possible.

PR #342, opened 2019-12-12
Add metosin/ring-http-response support to exception middleware.

mgrbyte:
Adds support for using `metosin/ring-http-response` functions with the exception-middleware.
Introduces a dependency on `metosin/ring-http-response`.


modules/reitit-middleware/src/reitit/ring/middleware/exception.clj
```diff
@@ -98,6 +99,7 @@
 (def default-handlers
   {::default default-handler
    ::ring/response http-response-handler
+   ::http-response/response http-response-handler
```
miikka: No need to add the dependency except for tests if you just spell out this keyword in full. Besides, for the dependency to actually work, it'd need to be added to `modules/reitit-middleware/project.clj`.
```suggestion
   :ring.util.http-response/response http-response-handler
```

mgrbyte: Thanks, I've made changes based on your suggestion.
I've moved the dependency on `metosin/ring-http-response` to the dev profile in project.clj - hope that's the right place.

miikka: Yeah, that's good.

PR #350, opened 2020-01-10
Update to latest ring-swagger-ui

heyarne:
This updates `ring-swagger-ui`, thereby updating `swagger-ui` to version 3. Note that `swagger-ui` 3.x [supports different versions of the OpenAPI spec](https://github.com/swagger-api/swagger-ui/tree/a24957e8ce938de162a05a02f166b3120b1a1b06) compared to the one used before. This should not be a problem because the `swagger.json` generated by `reitit-swagger` uses the `swagger` fromat `2.0`, at least in my case. I don't know if there are cases where this might cause a problem.

### DIFF
--- a/doc/ring/swagger.md
+++ b/doc/ring/swagger.md
@@ -63,6 +63,8 @@ Webjars also hosts a [version](https://github.com/webjars/swagger-ui) of the swa
 
 **NOTE**: Currently, swagger-ui module is just for Clojure. ClojureScript-support welcome as a PR!
 
+**NOTE:** If you want to use swagger-ui 2.x you can do so by explicitly downgrading `metosin/ring-swagger-ui` to `2.2.10`.
+
 ## Examples
 
 ### Simple example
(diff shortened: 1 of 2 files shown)


miikka:
I've intentionally kept not updating it because I like the version 2 more. But maybe we should move on with the times. :P

ikitommi:
Same here! One option would be to push the ui's into separate resource dirs, e.g. `/resources/swagger/2` and `/resources/swagger/latest`. Not suggesting, but an option.

heyarne:
> I've intentionally kept not updating it because I like the version 2 more. But maybe we should move on with the times. :P

What exactly did you like more about it? I had a problem with a swagger file created with `reitit-swagger` that caused an error with version 2 but not version 3, which is why I upgraded. I can't share the file here unfortunately, but maybe in private / the #clojurians slack.

ikitommi:
I tested with few apps, seems to work ok. Let's use the latest. One can always revert to older one if needed.

ikitommi:
Oh, but there is now a merge conflict. Could you please resolve that? Also, a mention to the docs on how to use the 2.* UI would be great.

heyarne:
Let me know if that works for you or if you need anything else!

ikitommi:
thanks!

PR #352, opened 2020-01-15
format negotiation documentation to reitit

wandersoncferreira:
As discussed in #reitit channel today, would be benefitial to improve the documentation towards how to change some options used by muuntaja to handle formatting, encoding and decoding of messages.

I thought about this first draft, very concise about what to do and how. Either some links to provide further investigation in the subject.


ikitommi:
Thanks for the PR! There was already a small part of content-negotiation in https://github.com/metosin/reitit/blob/master/doc/ring/default_middleware.md#content-negotiation

Suggestions: break down the "default-middleware" page into parts, one being the content-negotiation and merge the current and your texts into that. Also, the https://github.com/metosin/reitit/blob/master/doc/SUMMARY.md should be updated so the GitBook docs stay in sync.

ikitommi:
Awesome, thanks!!

PR #356, opened 2020-01-31
Add missing spec-tools dependency to reitit-middleware. Fixes #355

### LINKED ISSUE #355: reitit.ring.middleware.multipart requires spec-tools.core
Current version of Reitit-middleware does not depend on spec-tools, but the implementation of reitit.ring.middleware.multipart requires it. The workaround for the user is not too bad (just require spec-tools in project).

However, since Spec is not a requirement for using Reitit and Spec-tools is used for internal implementation of Reitit-middleware, Spec-tools should probably be required by Reitit-middleware and not force the user to depend on it.

MattiNieminen:


ikitommi:
Thanks for the PR.

Another way for this would be to push the spec definitions into `reitit-spec` module, under `reitit.ring.spec`, like both Schema and Malli have already:

* https://github.com/metosin/reitit/blob/master/modules/reitit-malli/src/reitit/ring/malli.cljc
* https://github.com/metosin/reitit/blob/master/modules/reitit-schema/src/reitit/ring/schema.cljc

would be a (small) breaking change, but for the better? 

what do you think?

MattiNieminen:
Probably for the better since it works the same way for Malli and Schema. Maybe this can be merged as a quick fix though?

miikka:
I think this should be merged

ikitommi:
true that, thanks!

PR #357, opened 2020-01-31
Update sieppari version to latest 0.0.0-alpha8

thenonameguy:


ikitommi:
thanks. this is a breaking change as core.async & manifold are not supported by default, need to document this and check how existing apps fail (fast/silently).

PR #359, opened 2020-02-02
Add janus to opensensors benchmark

aisamu:
@ikitommi gave me the 👍 on clojurians, so here it is!

[Janus](https://github.com/cch1/janus) is a zipper-based data-centric router for clj/s

There are some Rich-comment-forms after the routing table since the original repo is a little light on usage examples, but I'd be happy to remove them.

Speed-wise, it positioned itself between `calfpath-unroll` and `compojure`.

There might still be some squeezing potential re: the routing table conciseness and speed, but I believe all the low-hanging fruits were collected after a review from the library author.

Once again, thanks for the extraordinary work!

### DIFF
--- a/perf-test/clj/reitit/opensensors_perf_test.clj
+++ b/perf-test/clj/reitit/opensensors_perf_test.clj
@@ -10,6 +10,8 @@
             [compojure.core :refer [routes context ANY]]
             [calfpath.core :as cp]
             [calfpath.route :as cr]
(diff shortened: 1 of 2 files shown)


ikitommi:
thanks!

ikitommi:
Just re-run Janus: https://github.com/metosin/reitit/commit/3fe7764a331e607de0f2e2fd7c175725d55eb4bf

PR #360, opened 2020-02-05
Pass malli options to extract-parameter fn

olimsaidov:
Reitit swagger fails to generate swagger schema when using custom malli registry:

```clojure
["/api"
 {:swagger    {:id ::app}
  :coercion   (reitit.coercion.malli/create {:options {:registry @custom-registry}})
  ["/swagger.json"
   {:get (reitit.swagger/create-swagger-handler)}]}]
```


ikitommi:
thanks!

PR #361, opened 2020-02-05
Pass malli options to compile fn

olimsaidov:
Reitit coercion fails when using custom registry 

```clojure
["/api"
 {:coercion   (reitit.coercion.malli/create {:options {:registry @custom-registry}})}
 ...]
```

This is because, default value of `compile` option `malli.util/closed-schema` is not called with malli options. 

Requires https://github.com/metosin/malli/pull/171 to be merged


ikitommi:
thx3!

Issue #362, opened 2020-02-09
When there are coercion errors in multiple places, only one is returned.


aeriksson:
Currently, only one coercion error will be returned, even if there are errors in multiple places (e.g. in both body and path parameters). Furthermore, _which_ error is returned seems to depend on the order in which the coercers is evaluated (which is arguably arbitrary from the user's perspective).

It would probably be a better idea to instead run all coercers even if one fails, and merge the resulting maps.

This would require some modifications to the coercion error data format, because the current format contains an `:in` key that can't really be generalized to include multiple data sources. Example (for schema):
```json
{
    "coercion": "schema",
    "errors": {
        "x": "(not (integer? \"a\"))"
    },
    "in": [
        "request",
        "path-params"
    ],
    "schema": {
        "Keyword": "Any",
        "x": "Int"
    },
    "type": "reitit.coercion/request-coercion",
    "value": {
        "x": "a"
    }
}
```

My suggestion would be to change this to scope keys by the parameter location:
```json
{
    "body": {
        "coercion": "schema",
        "errors": "(not (map? nil))",
        "schema": {
            "z": "Int"
        },
        "value": null
    },
    "path-params": {
        "coercion": "schema",
        "errors": {
            "x": "(not (integer? \"a\"))"
        },
        "schema": {
            "Keyword": "Any",
            "x": "Int"
        },
        "value": {
            "x": "a"
        }
    },
    "query-params": {
        "coercion": "schema",
        "errors": {
            "y": "missing-required-key"
        },
        "schema": {
            "Keyword": "Any",
            "y": "Int"
        },
        "value": {}
    },
    "type": "reitit.coercion/request-coercion"
}
```

Above I've kept `"type"` in the top level, and scoped everything else by the parameter location, because it is by necessity the same regardless of where the error happens. Maybe the same should be done for `coercion`?

PS. While we're talking about the data format:
- Should the keys `schema`, `type`, and `coercion` really be included in the error response by default? Not sure that these are useful to end-users. Maybe there should be middleware/interceptor options to control this?
- Perhaps `value` for `query-params` should be `null` when no query params are provided? This would be more consistent with how `body` works, and probably more intuitive as well.

PR #363, opened 2020-02-09
Optionally coerce all params even if one fails.

aeriksson:
This PR targets issue #362.

This commit adds a `:reitit.coercion/coerce-all-on-error?` option that
ensures that all parts of requests are coerced even if one fails.
Previously, only the first coercion failure would be returned.

Since the current coercion exception ex-data format can't really be
used to accomodate multiple errors, this format changes from:
`{ "in": [ "request", "path-params" ], "errors": ... }`
..to:
`{ "path-params": { "errors": ...  }, "body": { "errors": ... } ...}`
when `:coerce-all-on-error?` has been set.

It probably makes sense to switch to the new format even when
`coerce-all-on-error?` is disabled, as well as and enable
`coerce-all-on-error?` by default at some point. However, it's probably
best to hold off on this until 1.0.0, since the format change means it
would break backwards compatability for downstream middleware and
interceptors.

I also made sure that response coercion behaves the same way as request
coercion wrt the new option, even though there's only one supported
response coercion type (body) currently.

I also added a missing `:` in the `::keys` option destructuring in
`response-coercer` (previously, users would have had to set
`:reitit.coercion/extract-request-format` for requests but
`:extract-response-format` for responses).

### DIFF
--- a/modules/reitit-core/src/reitit/coercion.cljc
+++ b/modules/reitit-core/src/reitit/coercion.cljc
@@ -41,7 +41,9 @@
    :header (->ParameterCoercion :headers :string true true)
    :path (->ParameterCoercion :path-params :string true true)})
 
-(defn ^:no-doc request-coercion-failed! [result coercion value in request]
+;; TODO: Fold this into `request-coercion-failed-join-errors!` in 1.0.0 when
+;;       switching to the new response format.
+(defn ^:no-doc request-coercion-failed-fail-fast! [result coercion value in request]
   (throw
     (ex-info
       (str "Request coercion failed: " (pr-str result))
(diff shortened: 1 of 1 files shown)


modules/reitit-core/src/reitit/coercion.cljc
```diff
@@ -83,42 +119,66 @@
...
+(defn encode-single-error [data]
+  (-encode-error (:coercion data) (update data :coercion -get-name)))
 
 (defn encode-error [data]
-  (-> data
-      (dissoc :request :response)
-      (update :coercion -get-name)
-      (->> (-encode-error (:coercion data)))))
+  (if (not-any? #(contains? data %)
+                [:body :form-params :query-params :headers :path-params])
```
aeriksson: This might not be the cleanest way to tell the old and new format apart. Perhaps it'd been better to handle this via an additional key to the `ex-info` data?

aeriksson:
Note: the tests seem to be failing on master, so I haven't actually run them on this branch. I can do a rebase, fix and force-push whenever master is working again.

modules/reitit-core/src/reitit/coercion.cljc
```diff
@@ -53,7 +55,27 @@
          :in [:request in]
          :request request}))))
 
-(defn ^:no-doc response-coercion-failed! [result coercion value request response]
+(defn ^:no-doc merge-coercion-errors
+  "Coalesces a list of coercion errors on the form `{:in <>}`"
```
aeriksson: Reminder to self: fix this docstring.

PR #365, opened 2020-02-13
-href on FragmentHistory should include the path

dspiteself:
If the path is not included and you create a link it will always point to 
#/$path
but your index file may not be root for instance
/version/test/index.html#/$path

### DIFF
--- a/modules/reitit-frontend/src/reitit/frontend/history.cljs
+++ b/modules/reitit-frontend/src/reitit/frontend/history.cljs
@@ -47,7 +47,7 @@
         fragment)))
   (-href [this path]
     (if path
-      (str "#" path))))
+      (str  (.. js/window -location -pathname) "#" path))))
 
 (defn- closest-by-tag [el tag]
   ;; nodeName is upper case for HTML always,


Deraen:
What kind of problems did you encounter and where?

Existing behaviour should work in the most common case, where `href` return value is used for anchor `href` attribute:

`<a href="#/items/2?foo=bar">Item 2</a>`

This works because if anchor `href` starts with hash, browser will only update the location hash, and keep path and others intact.

You can test this with frontend example, and opening e.g.  
http://localhost:3449/hello/world/, all the links still work because browser only changes the hash when clicking on the links.

dspiteself:
I guess the bug should be fixed in [push-state](https://github.com/metosin/reitit/blob/0.3.9/modules/reitit-frontend/src/reitit/frontend/history.cljs#L172) 
When you call push-state with fragment-history It will not respect your existing path.

Deraen:
I tested this locally on on examples/frontend, using the same url I mentioned in my previous comment. The index view contains buttons "item 3" and "replace state item 4" which call pushState and replaceState, and the both work correctly and keep the path intact and only update the fragment.

PR #369, opened 2020-02-28
Update dynamic_extensions.md

mchughs:
I find the clarity of this example important because the implementation of (ring/get-match) https://github.com/metosin/reitit/blob/c23f591283883bcebae482e723aa34284cfd048c/modules/reitit-ring/src/reitit/ring.cljc#L309 doesn't explain what's going on here.

I find the destructing of the qualified keyword `::roles` unnecessary to the example. To this end I propose using an unqualified keyword.

Additionally there is a collision between the name of the required rolls to access the route and the name of the roles held by the user making the request. To this end I propose having the roles held by the user named `:my-roles`


ikitommi:
thanks.

PR #371, opened 2020-03-05
Rename old reitit.ring.spec and move multipart specs to new ns

### LINKED ISSUE #355: reitit.ring.middleware.multipart requires spec-tools.core

Deraen:
This should fix #355 #356


ikitommi:
This is a major breaking change. I propose we'll postpone this to 1.0.0, could be soon (lot's of other... wrong things like the one this fixes.

miikka:
This could be handled in staged fashion to 

1. Move the specs to new a namespace from multipart.clj but add aliases for them in multipart.clj. Document that the aliases are deprecated (update code examples, add comment in code, entry in changelog. Nobody is looking at `^:deprecated` tags for specs, I guess.)
2. Later remove the aliases when doing a breaking release.

I'm not sure why `reitit.ring.spec` is renamed `reitit.ring.validate.spec` ?

PR #372, opened 2020-03-05
Ensure rfe/history is set during initial on-navigate call

### LINKED ISSUE #315: frontend.easy
Initial match doesn't have @history set when using reitit.frontend.easy, so push-state/replace-state fail.

```
(reitit.frontend.easy/start!
                 router
                 (fn [match history]
                   (push-state ::login))
                 {:use-fragment false})
```

Can be worked around by using reitit.frontend.history directly with provided history argument.

Deraen:
Fixes #315


ikitommi:
Awesome!

Issue #134, opened 2018-08-31
Parameters middleware


ikitommi:
leftover from #85 - currently `wrap-params` is required from Ring, could be optimized and packaged into `reitit-middleware`. the body-params could be pushed into Muuntaja (see https://github.com/metosin/muuntaja/issues/27).

Also, query-params could support the `:rails` format: https://github.com/JulianBirch/cljs-ajax#get-specific-settings

ikitommi:
initial version done.

green-coder:
It is possible to tag issues in commit messages (for example "related to #134"), Github will automatically add a link to the commit here.

valerauko:
I can help finish this if you linked the commit of that "initial version" you mention

ikitommi:
There is https://github.com/metosin/reitit/blob/master/modules/reitit-middleware/src/reitit/ring/middleware/parameters.clj, which is the current ring impl, I don't think there is anything else around. This issue is quite big, there could be 3 subissues:

- [x] the `parameters-middleware` & `paramters-interceptor`
- [ ] remove the `:form-param` thing and add a new format into Muuntaja instead -> it should be part of normal content-negotiation, not a separate step
- [ ] the query-param formats, e.g. `:rails`

devurandom:
I currently handle form encoded requests like this:

```clj
(require '[reitit.coercion.spec :as rcs])
(require '[reitit.ring :as reitit-ring])
(require '[reitit.ring.coercion :as rrc])
(require '[reitit.ring.middleware.parameters :as parameters])

(reitit-ring/ring-handler
  (reitit-ring/router
    [["//some/path" {:post {:swagger {:consumes ["application/x-www-form-urlencoded"]
                                      :produces ["text/plain"]}
                            :parameters {:form {:text string?}}
                            :responses {200 {:body string?}}
                            :handler my-handler
                            :coercion rcs/coercion
                            :middleware [parameters/parameters-middleware
                                         rrc/coerce-exceptions-middleware
                                         rrc/coerce-request-middleware
                                         rrc/coerce-response-middleware]}}]]))
```

I am dealing with a caller to my API that does not appear to be very strict when it comes to this, so I would like to stick to the documentation they provide instead of doing auto negotiation. Thus I would like to keep the current behaviour and assume a form-encoded request and force a plain-text response, independently of the presence of `content-type` and `accept` headers.

How would I do this with the new Muuntaja-based form-encoded body handling?

devurandom:
[...]

PR #381, opened 2020-04-06
Split parameters middleware into query and form + add string-transformer as form format

caioaao:


devurandom:
https://github.com/metosin/reitit/issues/134 appears to be the issue that covers the discussion of this.

PR #388, opened 2020-04-24
add optional key example for clojure spec

jaimesangcap:
Updated the docs for clojure spec to include example for optional keys


ikitommi:
Thanks!

PR #389, opened 2020-04-26
Quick router creation

ikitommi:
* Reuse path splitting results in path conflict resolution, O(n2) -> O(n) complexity
* common path-conflict resolution can be disabled by defining `:router` and setting `:conflicts` to nil. With 100 routes, a `r/linear-router` creation is ~15x faster (7ms vs 100ms)
  * e.g. `(r/router routes {:router r/linear-router, :conflicts nil})` forces a linear router, with no conflict checking (fast to create, slower to match)
* quarantine-router resolves path-conflicts just once (2x faster)

### With 100 conflicting routes, path conflict resolution 200ms -> 13ms

~~Still, the path conflict resolution is really slow. Here's a flamegraph:~~

Screenshot 2020-04-26 at 22 22 18

### DIFF
--- a/modules/reitit-core/src/reitit/core.cljc
+++ b/modules/reitit-core/src/reitit/core.cljc
@@ -301,7 +301,7 @@
   ([compiled-routes]
    (quarantine-router compiled-routes {}))
   ([compiled-routes opts]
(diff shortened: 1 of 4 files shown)


mk:
Can confirm that this improves our ClojureScript router init time from ~1s to ~100ms. Thank you!

Issue #403, opened 2020-05-19
[0.5.1] Swagger-UI doesn't show response body when no responses coercion.


piotr-yuxuan:
Once again, thanks for the great job you're doing here. It's really a pleasure to work this reitit+malli in 'real-life' scenarii.

Minimal description: on latest release `0.5.1` Swagger UI doesn't behave and hides the response when malli (baked-in version) coercion is used, but no coercion is applied on the response. An obvious workaround is provided below.

Example with good and bad routes:
```
["/some/path/not-working" {:get {:summary "minimal working example"
                                 :handler (constantly {:body {:message "This won't be shown in the UI"}})}}]
["/some/path/working" {:get {:summary "minimal working example"
                             :handler (constantly {:body {:message "This won't be shown in the UI"}})
                             :responses {200 {:body any?}}}}]
```

Would you accept a PR to fix this? Would you have any suggestion on how to deal with it? I would suggest adding `{200 {:body any?}}` as default responses coercion.

piotr-yuxuan:
Addendum: picture on the behaviour

<img width="957" alt="Screenshot 2020-05-19 at 15 45 35" src="https://user-images.githubusercontent.com/3686502/82342018-09426f80-99e9-11ea-8c2d-92a4320486ee.png">

ikitommi:
It seems that `:responses` is mandatory in the swagger spec. Adding a `:body` will cause a a response coercion, which slows things down. I suggested the following: if an endpoint doesn't have `:responses` defined, `:responses {:default {:description ""}}` will be added. It both makes the spec valid, enables showing the responses in swagger-ui and doesn't trigger the response coercion.

The ns that should do this is https://github.com/metosin/reitit/blob/master/modules/reitit-swagger/src/reitit/swagger.cljc. PR most welcome (test(s) included)

piotr-yuxuan:
Fixed upstream.

PR #404, opened 2020-05-19
[0.5.1] Ass default :responses for Swagger-UI

piotr-yuxuan:
https://github.com/metosin/reitit/issues/403

### DIFF
--- a/modules/reitit-swagger/src/reitit/swagger.cljc
+++ b/modules/reitit-swagger/src/reitit/swagger.cljc
@@ -92,6 +92,7 @@
                                      (apply meta-merge (keep (comp :swagger :data) interceptors))
                                      (if coercion
                                        (coercion/get-apidocs coercion :swagger data))
+                                     (when-not (:responses data) {:responses {:default {:description ""}}})
                                      (select-keys data [:tags :summary :description])
                                      (strip-top-level-keys swagger))]))
            transform-path (fn [[p _ c]]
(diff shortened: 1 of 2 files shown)


piotr-yuxuan:
Due to the position of `{:responses {:default {:description ""}}}` in the meta-merge args, one could argue that `when-not (:responses data)` is superfluous. That's probably true. On the other hand, I feel like it makes code intent more explicit.

modules/reitit-swagger/src/reitit/swagger.cljc
```diff
@@ -92,6 +92,7 @@
                                      (apply meta-merge (keep (comp :swagger :data) interceptors))
                                      (if coercion
                                        (coercion/get-apidocs coercion :swagger data))
+                                     (when-not (:responses data) {:responses {:default {:description ""}}})
```
ikitommi, line 95: I think this should be outside of the meta-merge: if the middleware or interceptors have defined the responses, this will override them. Or, add a meta-merge hint that this values loses on merge, if such a hint exists.

piotr-yuxuan: 6e332e46f1f7496bf94ceb42a415bfc4ef1b49ed is an attempt to build on your remark. I'm having hard time to see where to put this default spec if not at the root of `meta-merge`, but other options I'm not aware of might be preferable.

ikitommi:
Thanks!

PR #412, opened 2020-06-11
fix: coercion.malli use m/type

rschmukler:
Renames `m/name` to `m/type` per the breaking change introduced in malli
4880734d554511f16fee9c1d28a9d340c8b632c1.

Worth noting that this doesn't update the managed dependency in `project.clj` because a release with `m/type` hasn't been cut yet.


ikitommi:
thanks, will make a new snapshot of malli to align this.

Issue #375, opened 2020-03-11
Unclear where swagger tags should be placed


Deraen:
Swagger feature docstring (and spec?) places tags on route-data root:

https://github.com/metosin/reitit/blob/master/modules/reitit-swagger/src/reitit/swagger.cljc#L55

Examples place tags key under swagger:

https://github.com/metosin/reitit/blob/master/examples/ring-swagger/src/example/server.clj#L30

Both seem to generate proper swagger.json, but first one fails spec validation. Why? Spec seems to list tags as top-level key.

ikitommi:
You need to mount `swagger-feature` as a middleware|interceptor, which adds the valid keys to the route data for the routes `swagger-feature` effects

ikitommi:
e.g. all specs from mw/int `:spec` key accumulate to the route spec. Would be great to see the accumulated spec per endpoint.

ikitommi:
I think the right solution is to remove the top-level `:tags` completely and use path `[:swagger :tags]` instead.

Koura:
I know it would be a breaking change but what about this line: https://github.com/metosin/reitit/blob/9ca88d7165c97ec1c5bb87dd5a217bc58ad1de50/modules/reitit-swagger/src/reitit/swagger.cljc#L96

ikitommi:
oh, it a breaking. but the `:tags` has never been part of the route spec,
so this is still a bug fix. need to push a minor out with this.

opqdonut:
got fixed in #414

PR #414, opened 2020-06-24
Replace top-level `:tags` with `[:swagger :tags]` path in documentation

Koura:
Change `:tags` usage in documentation as mentioned in #375


ikitommi:
thanks!

PR #415, opened 2020-06-25
Cleanup test namespaces

Koura:
Make `default-options-handler-test` run by renaming the test and clean up duplicate require of `schema.core` in `test/cljc/reitit/swagger_test.clj`.


ikitommi:
thx!

PR #417, opened 2020-07-10
Issue 217 consume form params

### LINKED ISSUE #217: Automatically publish Swagger :consumes for :form params
Currently, on has to add manually `:swagger {:consumes ["application/x-www-form-urlencoded"]}` to endpoints that accept `:form` params. This can be automated:

* if the endpoint has `:form`, but not `:body` params, the [parameters-middleware](https://github.com/metosin/reitit/blob/master/modules/reitit-middleware/src/reitit/ring/middleware/parameters.clj) & [parameters-interceptor](https://github.com/metosin/reitit/blob/master/modules/reitit-interceptors/src/reitit/http/interceptors/parameters.clj) should publish the needed `:consumes`

example how to do that in the [muuntaja-middleware](https://github.com/metosin/reitit/blob/master/modules/reitit-middleware/src/reitit/ring/middleware/muuntaja.clj#L64)

#84 will outdate this. But a quick-win while waiting for that.

EDIT: I think the Muuntaja Middleware/Interceptor should also be changed that if there are is `:form`, but no `:body` parameters defined, it will not push it's own swagger-data. Otherwise, the order of the mw matters (the last one overrides), which would be awful.

Koura:
Automatically publish `:swagger {:consumes ["application/x-www-form-urlencoded"]}` for parameters middleware/interceptor namespaces when endpoint has `:form` but no `:body` parameters present.

In similar situation omit swagger-data for muuntaja middleware/interceptor to avoid problems with middleware ordering.

Closes #217

### DIFF
--- a/modules/reitit-interceptors/src/reitit/http/interceptors/muuntaja.clj
+++ b/modules/reitit-interceptors/src/reitit/http/interceptors/muuntaja.clj
@@ -9,6 +9,16 @@
 (defn- displace [x] (with-meta x {:displace true}))
 (defn- stripped [x] (select-keys x [:enter :leave :error]))
 
+(defn- publish-swagger-data? [{:keys [form body]}]
+  (not (and (some? form)
+            (nil? body))))
+
+(defn- swagger-data [parameters muuntaja]
+  (if (publish-swagger-data? parameters)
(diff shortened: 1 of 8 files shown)


ikitommi:
missed this, looks great, thanks!

PR #419, opened 2020-07-11
Add Buddy example

vharmain:
This PR adds an annotated example how to use [Buddy](https://github.com/funcool/buddy) to implement simple authentication and authorization flows with reitit.


miikka:
I added some curl examples in the README as well: https://github.com/metosin/reitit/commit/486e3928c5e50d72281619cec01ddd2240ca6adb

PR #427, opened 2020-08-04
Fix: broken ring-malli-swagger example

Koura:
Ring-malli-swagger example doesn't compile because `malli.core/accept` and `malli.core/schema-visitor` aren't recognized anymore. Remove redundant code block in order to make the example compile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ pom.xml.asc
 /_book
 figwheel_server.log
 /.idea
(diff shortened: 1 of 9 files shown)


Koura:
> Thanks, the whole code block seems redundant, could you please remove it?

Good point :+1: I was too fixated on getting it to work to notice that

ikitommi:
thanks!

PR #428, opened 2020-08-17
Fix rfe/start! example arity

nivekuil:
The second arg should be the on-navigate callback and the opts should be the third arg.


ikitommi:
thanks!

PR #439, opened 2020-10-01
Fixing broken links to documentation in README

nate:
This tripped me up just now when I was looking at the project and it seemed a quick thing to fix. Cheers!


ikitommi:
thanks!

PR #442, opened 2020-10-14
Update nested controllers example, :params -> :parameters

viesti:
I noticed that the nested controllers example still uses the deprecated `:params` form, so let's update it to use `:parameters` form.


ikitommi (approved):

ikitommi:
thanks!!!

PR #448, opened 2020-11-01
Link to docs for Malli coercion

teodorlu:
How to do request coercion with Malli is documented in the doc/ folder, but the docs don't show up on cljdoc. This PR adds a link to the Malli coercion docs so that cljdoc might index those documents.

I haven't been able test cljdoc ingestion locally, so I'd appreciate a second pair of eyes.


ikitommi:
thanks!

PR #449, opened 2020-11-02
Add route-data to coerce-request compile failure

Deraen:


modules/reitit-ring/src/reitit/ring/coercion.cljc
```diff
@@ -24,24 +24,33 @@
...
+                      (fn
+                        ([request]
+                         (let [coerced (coercion/coerce-request coercers request)]
+                           (handler (impl/fast-assoc request :parameters coerced))))
+                        ([request respond raise]
+                         (let [coerced (coercion/coerce-request coercers request)]
+                           (handler (impl/fast-assoc request :parameters coerced) respond raise)))))
+                    {})
+                  (catch Exception e
+                    (throw (ex-info (str "Coerce-request-middleware failed: " (.getMessage e))
```
ikitommi, line 47: `reitit/exception/fail!` could be used instead?

Deraen: It doesn't currently take cause as parameter.

PR #452, opened 2020-11-13
Support operationId in reitit-swagger

### LINKED ISSUE #451: Support operationId in reitit-swagger
Hi,
thanks for creating this. I wonder if there is something to be said against supporting `operationId` as swagger element. I could use it right now for automated client creation on a spec from server.

TimoKramer:
OpenAPI Specification 3.0 allows the operationId to be added to the
"Operation Object" alongside e.g. summary and description. Since
this is useful in some cases like auto-generating clients, it should be
supported. This commit introduces the support of this element in the
reitit-swagger module and extends the test with an operationId in
two places.

- Spec: https://swagger.io/specification/#operation-object
- Adds the support for operationId
- Adds operationId in two places of the swagger test
- Closes #451


TimoKramer:
A friend of mine pointed out, that operationId is unique. So there should be some kind of error handling if the spec is bad.

sundbry:
IMHO the openapi spec validation is ok to leave to outside tools. If the user defines a non unique one thats their mistake like any other programing error. Its good enough we can at least have this capability! Thank you @TimoKramer 

Should there be support for kebab-case `:operation-id`?

evilsneer:
Hi! Can I hope for merge soon?

miikka:
Note that the Swagger support in Reitit generates Swagger, not OpenAPI 3.0. However, Swagger too has `operationId` ([ref](https://swagger.io/docs/specification/2-0/paths-and-operations/#operationid)) so we could still have this change.

miikka:
Sorry for taking such a long time to review. I pushed a change to simplify this patch to make it mergeable.

* I think that we can leave out the uniqueness check – we do not really validate the Swagger output anyway.
* Regarding the proposed implementation for the check, I'd prefer to separate it from the transformation code. Not a big fan of using atoms to gather information during the transform unless it has big performance benefits.

This needs to be also documented in https://cljdoc.org/d/metosin/reitit/0.5.12/doc/ring/swagger-support (but that can be a separate PR).

The question about `:operationId` vs `:operation-id` is good. Since this is a top-level key and not under `:swagger`, it could be argued that it should be kebab-case to be inline with the other keys. On the other hand, it'd be great to follow closely the naming in Swagger specification.

TimoKramer:
Any decision on `operationId` vs `operation-id`?

tommy-mor:
I would like this feature, any update on this?

evilsneer:
just make it kebab-case and release already!

ikitommi:
Thanks!!

PR #456, opened 2020-12-01
Fix relative links in documentation

raphaelsaunier:
This PR fixes two broken internal links in the documentation and a link to a source file.

~There's also a broken external link to http://thinkrelevance.com/blog/2013/06/04/clojure-workflow-reloaded. I've notified Cognitect and @stuartsierra; based on their reply we can either replace or remove the link altogether.~

_Update Dec 2, 2020: the article is now accessible again. Thank you @admiralbumblebee!_

I ran the whole `doc/` folder through [`markdown-link-check`](https://github.com/tcort/markdown-link-check) and found no other broken links.


ikitommi:
thanks!

PR #458, opened 2020-12-15
Downcase hidden methods in RESTful example

deobald:
- this documentation is mildly confusing when combined with hiccup's
  `form-to` since hiccup forcibly transforms the method specified in
  its convenience syntax: `(form-to [:delete "/my-url"] ... )` into an
  upper-case string:
  https://github.com/weavejester/hiccup/blob/80e48352dd2bce5277ccdd06334cfb9be915afb4/src/hiccup/form.clj#L130

- it's also common to get an upper-case string from elsewhere so it
  seems best to wrap the hidden `_method` in `lower-case`.


ikitommi:
thanks!

Issue #464, opened 2021-01-21
reitit.ring/create-resource-handler :not-found-handler is not called 


paulbutcher:
I’m unable to get the `:not-found-handler` option for `reitit.ring/create-resource-handler` to work: it doesn’t look like the function is called at all.

I’ve created a minimal cut-down version of my application which demostrates the problem: https://github.com/paulbutcher/reitit-bug

The relevant code is:

```clojure
(def handler
  (ring/ring-handler
   (ring/router
    ["/ping" {:get (constantly (response "pong"))}])
   (ring/routes
    (ring/create-resource-handler {:path "/"
                                   :not-found-handler (fn [x]
                                                        (println "not found handler: " x)
                                                        (-> "index.html"
                                                            (resource-response {:root "public"})
                                                            (content-type "text/html")))})
    (ring/create-default-handler))))
```

miikka:
The not found handler only works in the case that `:path` has not been set as can seen from the implementation, linked below. I'm not sure about the rationale behind the implementation.

https://github.com/metosin/reitit/blob/0ecc2fb9a7c5d93f2bb7582fbdedf75e8dc8e484/modules/reitit-ring/src/reitit/ring.cljc#L221-L230

My guess is that it was left out because `:not-found-handler` defaults to a 404 handler and not to `(constantly nil)`, preventing the handler from composing the usual way (which is what you want when `:path` is set). It think it could make sense to change it so that if `:path` matches but the resource is not found, then the not found handler is called.

To make the change backwards compatible, the default value for `:not-found-handler` would need to be set to `(constantly nil)` in case `:path` is set.

miikka:
Will be fixed in the next release. https://github.com/metosin/reitit/pull/473 needs to resolved first.

miikka:
Released in 0.5.13.

PR #473, opened 2021-02-26
Make the not-found-handler fix backwards compatible

miikka:
PR #471 aimed to fix issue #464. However, the change was slightly backwards-incompatible, since it made the file and resource handlers use the default 404 handler when mounted outside of the router. The previous behavior was to return nil in that case.

This patch restores the previous behavior and clarifies that `:path` option can be used only when the file/resource handler is mounted outside of a router.

I'm not completely sure if the previous behavior is good. Maybe we should make a minor breaking change here instead of fixing this?

---

TODO:

* [x] Update the user guide as well https://cljdoc.org/d/metosin/reitit/0.5.12/doc/ring/static-resources

### DIFF
--- a/modules/reitit-ring/src/reitit/ring.cljc
+++ b/modules/reitit-ring/src/reitit/ring.cljc
@@ -195,7 +195,9 @@
                         root "public"
                         index-files ["index.html"]
                         paths (constantly nil)
-                        not-found-handler (constantly {:status 404, :body "", :headers {}})}}]
+                        not-found-handler (if path
+                                            (constantly nil)
+                                            (constantly {:status 404, :body "", :headers {}}))}}]
      (let [options {:root root
                     :loader loader
                     :index-files? false
@@ -239,7 +241,7 @@
      | -------------------|-------------|
      | :parameter         | optional name of the wildcard parameter, defaults to unnamed keyword `:`
      | :root              | optional resource root, defaults to `\"public\"`
-     | :path              | optional path to mount the handler to. Works only if mounted outside of a router.
+     | :path              | path to mount the handler to. Required when mounted outside of a router, does not work inside a router.
      | :loader            | optional class loader to resolve the resources
      | :index-files       | optional vector of index-files to look in a resource directory, defaults to `[\"index.html\"]`
      | :not-found-handler | optional handler function to use if the requested resource is missing (404 Not Found)"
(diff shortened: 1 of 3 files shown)


miikka:
> I'm not completely sure if the previous behavior is good. Maybe we should make a minor breaking change here instead of fixing this?

Upon further thought, in general I'd expect Ring handlers to return nil in a 404 type situation. Thus I think we should continue with the previous behavior.

modules/reitit-ring/src/reitit/ring.cljc
```diff
@@ -239,7 +241,7 @@
      | -------------------|-------------|
      | :parameter         | optional name of the wildcard parameter, defaults to unnamed keyword `:`
      | :root              | optional resource root, defaults to `\"public\"`
-     | :path              | optional path to mount the handler to. Works only if mounted outside of a router.
+     | :path              | path to mount the handler to. Required when mounted outside of a router, does not work inside a router.
      | :loader            | optional class loader to resolve the resources
      | :index-files       | optional vector of index-files to look in a resource directory, defaults to `[\"index.html\"]`
      | :not-found-handler | optional handler function to use if the requested resource is missing (404 Not Found)"
```
ilmoraunio, line 247: I wonder if this is misleading (I am not sure) when it seems that a 404 Not Found may not always be returned... or will it? Applies to both documented instances of `:not-found-handler`.

ilmoraunio (commented):
Flush comment

modules/reitit-ring/src/reitit/ring.cljc
miikka: Here's how it is supposed to work: if `:not-found-handler` is set and the request is to a path that could be served by the file/resource handler (inside router, any request it gets; outside router, requests that match `:path`), but the file/resource is not found, then not-found-handler gets called.

I think this docstring is okay. Not sure how to improve it, but willing to take suggestions.

ilmoraunio: So, do we still return a 404 Not Found for requests coming from outside the router (ie. when `:path` is defined)? If so, then omission of `(404 Not Found)` would suffice.

miikka: If you give this parameter and set it to a handler that returns 404, then yes 😛

ilmoraunio: Well, in any case, this is more like a [nit] and we can address this as a separate PR if it truly is a concern for anyone.

ilmoraunio (approved):
Thanks!

miikka:
The failing ClojureScript build is not caused by this PR (it was broken by the commit https://github.com/metosin/reitit/commit/0c93cfb1f964d655b2063e5c6b94bf2128ccea28), so I'm merging this.

PR #480, opened 2021-03-11
Fix exception log handler, *out* is Writer, not PrintWriter

Deraen:
`*out*` is usually OutputStreamWriter (set by clojure.lang.RT)
`*out*` could also be java.io.StringWriter (with-out-str)

TODO: Needs tests


ikitommi:
"TODO: Needs tests"

Deraen:
Merged in https://github.com/metosin/reitit/pull/481

PR #487, opened 2021-04-22
Fix Ring example

audriu:
Maybe it is not perfect solution but at least it works. It was somewhat weird that the front page of the lib contains non working examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ A Ring routing app with input & output coercion using [data-specs](https://githu
               :muuntaja   m/instance
               :middleware [parameters/parameters-middleware
                            rrc/coerce-request-middleware
+                           rrc/coerce-response-middleware
                            muuntaja/format-response-middleware]}})))
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,24 +85,28 @@ Optionally, the parts can be required separately.
...
+                                     {:status 200
+                                      :body   {:total (+ x y)}})}}]]
       ;; router data affecting all routes
-      {:data {:coercion reitit.coercion.spec/coercion
-              :middleware [rrc/coerce-exceptions-middleware
+      {:data {:coercion   reitit.coercion.spec/coercion
+              :muuntaja   m/instance
+              :middleware [parameters/parameters-middleware
                            rrc/coerce-request-middleware
-                           rrc/coerce-response-middleware]}})))
```
ikitommi, line 105: this is needed too, to coerce responses.

ikitommi:
thanks!

PR #488, opened 2021-04-24
Update content_negotiation.md

rlovtangen:
Not 100% sure, but looks like it should be :date-format, not :data-format here, based on the example in the section above.


miikka (approved):
Oh yes. Those are basically the options used by jsonista's object-mapper constructor https://cljdoc.org/d/metosin/jsonista/0.3.2/api/jsonista.core#object-mapper

Thanks!

PR #495, opened 2021-06-24
Enrich request for pedestal/routing-interceptor default-queue

kommen:
This ensures requests handled by the `default-queue` in `pedestal/routing-interceptor`
also have access to the router per the injected `:reitit.core/router` key on the request.

Without this, interceptors in the default queue don't have access to the router.

For example, this is needed for the [`trailing slash handler`](https://github.com/metosin/reitit/blob/de5bdeba192c6d016af20feada18243ae7bcb4c0/doc/ring/slash_handler.md), which [relies on the router being injected](https://github.com/metosin/reitit/blob/de5bdeba192c6d016af20feada18243ae7bcb4c0/modules/reitit-ring/src/reitit/ring.cljc#L137).

The commit contains a test which fails without the change in `reitit.http`.


ikitommi:
Thanks!

PR #499, opened 2021-07-27
improve some docs

branch14:
Just ironed out some wrinkles while reading the docs.


branch14:
Oops, didn't mean to mess with the indentations like that. :(

PR #501, opened 2021-08-03
Fix malli encoding & update deps

### LINKED ISSUE #498: Malli response coercision seems to do nothing at all
Maybe I'm crazy, but after much debugging, I became convinced Malli response coercion doesn't do anything.

Evidence:
Looking at [this line](https://github.com/metosin/reitit/blob/master/modules/reitit-malli/src/reitit/coercion/malli.cljc#L51), if no `encoder` parameter is passed, the encoding simply returns the value - [see here](https://github.com/metosin/reitit/blob/master/modules/reitit-malli/src/reitit/coercion/malli.cljc#L72).

Now, looking at how response coercion is defined [here](https://github.com/metosin/reitit/blob/master/modules/reitit-malli/src/reitit/coercion/malli.cljc#L180-L183), it seems that the returned coercer uses another coercer as its `encoder` which, in turn, gets `nil` as its encoder - which means nothing is ever encoded.

This is also what seems to happen in practice. I can provide more details about my use case, but it seems like it doesn't really matter, given this code.

Thanks.

ikitommi:
Fixes #498


ikitommi:
yes, the default is used so that the idempotent defaults are run before. I think it's a bad idea, but didn't want to break the old behavior. Extra decoding-step Could be removed in 1.0.0?

PR #506, opened 2021-08-30
Faster keywordize

bsless:
Implement walk with protocols and keywordize keys with it.
Gives about 2x speedup

left to do: decide on an implementation for cljs


bsless:
@ikitommi do you want to extend this solution to cljs or keep it just for clj?

bsless:
Decided to split the implementation off only for Clojure since cljs is built differently, don't even know how `walk` is implemented for it

bsless:
@ikitommi are there any blocking items for a review? I'd be happy to clear them if so.

bsless:
Open question - how can we be sure this does not introduce regressions for users who upgrade?

bsless:
Added generative test to validate my assumption

ikitommi:
Thanks!

PR #512, opened 2021-10-06
Add a test for `/{foo}.html` style routes

miikka:
Just like path `///` would not match `/{foo}/`, `/..html` does not match `/{foo}.html`.


ikitommi:
true that!

PR #514, opened 2021-10-13
duplicate r/routes output

zengxinhui:
Perhaps this is not needed as the same output is shown above close.
Or the `nil`s need to be removed to match actual output. See PR513 https://github.com/metosin/reitit/pull/513


Deraen:
Merged: https://github.com/metosin/reitit/commit/153eaaf2b0739d76cede53a00e16a75ed4c4775d

PR #518, opened 2021-11-08
Fix middleware ordering in ring example

15joeybloom:
As is, the example ring app gives me response coercion errors because the response body is a stream rather than a map.


ikitommi:
thx!

PR #521, opened 2021-12-04
Remove unused internal ‘goog-extend’ function

mthl:
This function is internal (meaning in a `*.impl` namespace) and not used anymore.


ikitommi:
thanks!

PR #535, opened 2022-03-02
Use namespace alias to refer Record types in schema coercion ns

Deraen:
Since Cljs 1.10.439 using qualified name hasn't worked to refer to
record types.

### DIFF
--- a/modules/reitit-schema/src/reitit/coercion/schema.cljc
+++ b/modules/reitit-schema/src/reitit/coercion/schema.cljc
@@ -26,11 +26,11 @@
    (fn [x]
      (cond
        #?@(:clj [(class? x) (.getName ^Class x)])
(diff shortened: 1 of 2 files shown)


ikitommi:
Is there a reason for it not working on latest CLJS?

Deraen:
The change is probably not necessary here.

The problem on the schema-tools side was:

```
(defn maybe? [schema]
  (instance? schema.core.Maybe schema))
``` 

And in such a case, ClojureScript can't know if the `schema` part of `schema.core.Maybe` refers to the local binding or namespace, and the local name wins, in that case, the value doesn't have `.core.Maybe` property. The recommendation is that dotted qualified names shouldn't be dependent upon on ClojureScript, accessing all types through namespace is preferred.

PR #537, opened 2022-03-05
Save three seq constructions

frwdrik:
If this change is interesting, I can also include a similar rewrite of `reitit.ring.spec/validate-route-data`.


ikitommi:
thanks!

PR #557, opened 2022-07-19
Polish pedestal chains when printing context diffs

handerpeder:
`print-context-diffs` includes the interceptor stack and queue when using pedestal. remove the pedestal variant of the keys in `polish`.


ikitommi:
Thanks!

Issue #422, opened 2020-07-16
malli parameters coercion merging broken


xificurC:
```clj
  (do
    (require '[reitit.ring :as ring] '[reitit.ring.coercion :as coercion] 'reitit.coercion.malli)
    (ring/router ["/{a}" {:parameters {:path [:map [:a string?]]}}
                  ["/{b}" {:parameters {:path [:map [:b string?]]}}]]
                 {:data {:coercion reitit.coercion.malli/coercion
                         :middleware [coercion/coerce-request-middleware]}}))
```

```

  Show: Project-Only All 
  Hide: Clojure Java REPL Tooling Duplicates  (11 frames hidden)

1. Unhandled clojure.lang.ExceptionInfo
   nth not supported on this type: Keyword
   #:reitit.exception{:cause #error {
    :cause "nth not supported on this type: Keyword"
    :via
    [{:type java.lang.UnsupportedOperationException
      :message "nth not supported on this type: Keyword"
      :at [clojure.lang.RT nthFrom "RT.java" 991]}]
    :trace
    [[clojure.lang.RT nthFrom "RT.java" 991]
     [clojure.lang.RT nth "RT.java" 940]
     [malli.core$_parse_entry_syntax$_parse__13937 invoke "core.cljc" 270]
     [malli.core$_parse_entry_syntax$fn__13946 invoke "core.cljc" 273]
     [clojure.core$mapv$fn__8445 invoke "core.clj" 6912]
     [clojure.lang.ArrayChunk reduce "ArrayChunk.java" 63]
     [clojure.core.protocols$fn__8154 invokeStatic "protocols.clj" 136]
     [clojure.core.protocols$fn__8154 invoke "protocols.clj" 124]
     [clojure.core.protocols$fn__8114$G__8109__8123 invoke "protocols.clj" 19]
     [clojure.core.protocols$seq_reduce invokeStatic "protocols.clj" 31]
     [clojure.core.protocols$fn__8146 invokeStatic "protocols.clj" 75]
     [clojure.core.protocols$fn__8146 invoke "protocols.clj" 75]
     [clojure.core.protocols$fn__8088$G__8083__8101 invoke "protocols.clj" 13]
     [clojure.core$reduce invokeStatic "core.clj" 6828]
     [clojure.core$mapv invokeStatic "core.clj" 6903]
     [clojure.core$mapv invoke "core.clj" 6903]
     [malli.core$_parse_entry_syntax invokeStatic "core.cljc" 273]
[...]

PR #561, opened 2022-08-11
add `:meta-merge-fn` option

pfeodrippe:
It's an option for those who have https://github.com/metosin/reitit/issues/422 without having to enforce anything to the user.

`meta-merge-fn` follows the same signature of `meta-merge.core/meta-merge`.


pfeodrippe:
What do you think, @ikitommi?

ikitommi:
looks good, I'll rename it to `:meta-merge` to follow the current convention of functions. Thanks!

PR #566, opened 2022-10-12
Handle empty seq as empty string in `query-string`

tvaisanen:
example:

instead of

```clojure
(query-string {:nil nil
               :vec []
               :seq-empty '()})
;; => "nil=&&"
```

now

```clojure
(query-string {:nil nil
               :vec []
               :seq-empty '()})
;; => "nil=&vec=&seq-empty="
```


ikitommi:
Thanks!

PR #569, opened 2022-11-14
Swagger: support operationId in generated swagger json

jcdlbs:
It's desirable to support "operationId" in the generated Swagger/OpenAPI JSON for tooling that automatically builds user-friendly HTTP clients against the JSON. The Clojure "Martian" project, for example, parses the JSON and excludes any API calls that don't have the operationId attribute since it would be difficult to build a meaningful human readable name from the verb, path, and parameters.

This patch takes a simple approach and adds operationId support in the same way that description is supported in the Swagger code. The only undesirable aspect to this patch that I see is that "operationId" is camel case in the Clojure code but hyphenating it and camel-casing it on the output didn't seem worth the added complexity.

Hope this pull request is acceptable and thanks for maintaining such a great project!


ikitommi:
Thanks!!

PR #576, opened 2022-12-06
Coercion errors perf

### LINKED ISSUE #575: Default printing when coercion fails is very expensive
While reitit is fast in its happy path, when coercion fails there are several paths which are extremely wasteful:

In coercion.cljc:

```clojure
(defn ^:no-doc response-coercion-failed! [result coercion value request response]
  (throw
   (ex-info
    (str "Response coercion failed: " (pr-str result))
    (merge
     (into {} result)
     {:type ::response-coercion
      :coercion coercion
      :value value
      :in [:response :body]
      :request request
      :response response}))))
```

`pr-str` serializes the entire result which might be very large

In reitit/coercion/malli

```clojure
       (-encode-error [_ error]
         (cond-> error
           (show? :humanized) (assoc :humanized (me/humanize error {:wrap :message}))
           (show? :schema) (update :schema edn/write-string opts)
           (show? :errors) (-> (me/with-error-messages opts)
                               (update :errors (partial map #(update % :schema edn/write-string opts))))
```
returning the schema and errors by default and serializing them to edn string is also very slow and expensive.

Should these be off by default?

bsless:
Closes #575

### DIFF
--- a/modules/reitit-core/src/reitit/coercion.cljc
+++ b/modules/reitit-core/src/reitit/coercion.cljc
@@ -47,13 +47,15 @@
     (if serialize-failed-result
       (str "Request coercion failed: " (pr-str result))
       "Request coercion failed")
-    (merge
-     (into {} result)
-     {:type ::request-coercion
-      :coercion coercion
-      :value value
-      :in [:request in]
-      :request request}))))
+    (-> {}
+        transient
+        (as-> $ (reduce conj! $ result))
+        (assoc! :type ::request-coercion)
+        (assoc! :coercion coercion)
+        (assoc! :value value)
+        (assoc! :in [:request in])
+        (assoc! :request request)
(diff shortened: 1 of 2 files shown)


bsless:
@ikitommi Sorry for bumping this issue, but I worry our servers could fall over too easily in this state, especially as our schema grows. Is there someone who can look at it?

ikitommi:
Oh, it was also on the exception message. Thanks for the fix!

bsless:
Thanks
I'll prepare a docs update regarding the breaking change and new options

Snurppa:
Was there info about the breaking change somewhere? Did not realize the possible breakage from the changelog, but I think we bumped into this as we used `:errors` to do some "custom humanization".

Nevertheless, this is a good improvement, thank you Ben!

PR #581, opened 2023-01-18
Add reitit-frontend support for fragment string

Deraen:
Reworks #554 to a more generic version.

It is enough that we read the fragment as a string. Users can use Malli coercion with `:decode/string` to decode the fragment values e.g. for the OAuth use-case.

There shouldn't be a need for Reitit to provide a function to split those values either, OAuth RFC specifies `application/x-www-form-urlencoded` format, which is supported by `ring.util.codec/decode-form` on Clojure and e.g. [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) on JS environments.

Malli test case shows how to decode the string to a map.

Blocked by [malli](https://github.com/metosin/malli/pull/818)

### DIFF
--- a/modules/reitit-frontend/src/reitit/frontend.cljs
+++ b/modules/reitit-frontend/src/reitit/frontend.cljs
@@ -1,5 +1,6 @@
 (ns reitit.frontend
   (:require [clojure.set :as set]
+            [clojure.string :as str]
             [reitit.coercion :as coercion]
             [reitit.core :as r])
   (:import goog.Uri
@@ -20,6 +21,19 @@
          (map (juxt keyword #(query-param q %)))
          (into {}))))
 
+(defn fragment-params
+  "Given goog.Uri, read fragment parameters into Clojure map."
+  [^Uri uri]
+  (let [fp (.getFragment uri)]
+    (if-not (seq fp)
+      {}
+      (into {}
+            (comp
+              (map #(str/split % #"="))
+              (map (fn [[k v]]
+                     [(keyword k) v])))
+            (str/split fp #"&")))))
+
 (defn match-by-path
   "Given routing tree and current path, return match with possibly
   coerced parameters. Return nil if no match found.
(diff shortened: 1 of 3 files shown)


ikitommi:
`[metosin/malli "0.10.1"]` should have the fix.

test/cljs/reitit/frontend/core_test.cljs
```diff
@@ -22,8 +32,10 @@
                 :path-params {}
                 :query-params {}
                 :path "/"
+                :fragment nil
                 :parameters {:query {}
-                             :path {}}})
+                             :path {}
+                             :fragment nil}})
```
ikitommi, line 38: others default to `{}`, but fragment defaults to `nil`. `{}` would be a consistent default.

test/cljs/reitit/frontend/core_test.cljs
```diff
@@ -99,21 +118,25 @@
...
         (is (= (r/map->Match
                  {:template "/:id"
                   :path-params {:id "5"}
                   :query-params {:mode "foo"}
                   :path "/5"
+                  :fragment "fragment"
                   :parameters {:path {:id 5}
-                               :query {:mode :foo}}})
+                               :query {:mode :foo}
+                               :fragment "fragment"}})
```
ikitommi, line 139: oh, I see. Like the `:body` in backend. Is this coherent to what we do with `:query` in case of not passing in value? e.g. what is emitted on `?kikka#kukka`

test/cljs/reitit/frontend/core_test.cljs
Deraen: It is a string, not a map. Could default to an empty string, though.

test/cljs/reitit/frontend/core_test.cljs
Deraen: No, the ring will consider query-string without `=` a single parameter without a value:
`{:query-params {:kikka ""} :fragment "kukka"}`

modules/reitit-frontend/src/reitit/frontend.cljs
```diff
@@ -37,12 +37,17 @@
                    coercion/coerce!)]
      (if-let [match (r/match-by-path router (.getPath uri))]
        (let [q (query-params uri)
-             match (assoc match :query-params q)
+             fragment (when (.hasFragment uri)
```
Deraen, line 40: If we want to have a fragment value as a string always, this condition could be dropped.
`hasFragement` just checks if the parsed value is a non-empty string; there is no different between URL without `#` and one with a fragment but nothing after `#`.

ilmoraunio (approved):

PR #582, opened 2023-01-19
Include query-string as part of location header in redirect-trailing-slash-handler

### LINKED ISSUE #565: redirect-trailing-slash-handler strips query-params
When using `redirect-trailing-slash-handler` the redirect URL is missing the query-params, e.g. something like`/foo?bar=1` becomes `/foo/`.

ilmoraunio:
Fixes #565


Deraen (approved):

ikitommi:
Thanks!

PR #585, opened 2023-01-31
Allow var handlers

djblue:
I typically like to use vars for handlers to allow for interactive development. Changing this spec worked for me locally. If this isn't desired behavior upstream, no worries 💯


opqdonut:
Good idea! Will need docs & examples.

opqdonut:
Thanks :tada:

PR #586, opened 2023-02-01
Fix incorrect ring-router doc references

bplubell:
It looks like documentation references to `ring-router` are left-overs from early README examples - I couldn't find any code that ever used the name. The Ring router named `reitit.ring/router`.

While it didn't take too long for me to realize why my `ring-router` was not working, I had to look the examples to figure out the name of the function was just `router` - which was confusing since the section header stated `reitit-router`.

It would be good to have someone knowledgeable with the docs review for possible breaking urls - specifically, does the change in `doc/cljdoc.edn` affect the generated `.../ring/ring-router` path? A link from `README.md` still references `.../ring/ring-router`.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -110,9 +110,9 @@ Reverse-routing:
 ;        :path "/api/orders/2"}
 ```
 
-## Ring-router
+## Ring router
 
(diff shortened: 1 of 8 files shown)


ikitommi:
Thanks!!

PR #563, opened 2022-09-05
 Support for OpenAPI3

souenzzo:
- [x] Schema -> OpenAPI3
- [x] Malli -> OpenAPI3 (using json-schema)
- [x] Spec -> OpenAPI3
- [x] Coercion for all content-types
- [x] Coercion per content-type

Fixes #84

PR #588, opened 2023-03-03
Initial Openapi3 support

opqdonut:
For #84

Replaces #563

This should be a minimum releasable version of openapi3 support now.

Missing bits which I'll address in future PRs:
- integrate more with muuntaja (get default content type etc)
- spec coercion doesn't generate `:additionalProperties false` unlike malli/schema
- multipart support


opqdonut:
I think this is now ready for merge. The PR introduces new features, but doesn't break old ones. Openapi support is marked as alpha.

There's still a some refactoring to do, and some ease-of-use features (like proper muuntaja integration). I bet users will also discover bugs.

ikitommi:
I'll look into this early next week.

modules/reitit-core/src/reitit/coercion.cljc
```diff
@@ -98,17 +111,24 @@
...
+                                         [format (-response-coercer coercion schema)])
+                                       (filter second)
+                                       (seq)
+                                       (into {}))
+          default (when body (-response-coercer coercion body))]
+      (when (or per-format-coercers default)
+        (fn [request response]
+          (let [format (extract-response-format request response)
+                value (:body response)
+                coercer (get per-format-coercers format (or default (fn [value _format] value)))
```
ikitommi, line 127: the fn `(fn [value _format] value)` is used in both request & response coercion and is generated per each request, could be extracted into a named fn in this ns.

modules/reitit-openapi/src/reitit/openapi.cljc
```diff
@@ -0,0 +1,118 @@
...
+           accept-route (fn [route]
+                          (-> route second :openapi :id (or ::default) (trie/into-set) (set/intersection ids) seq))
+           ;base-openapi-spec {:responses ^:displace {:default {:description ""}}}
+           transform-endpoint (fn [[method {{:keys [coercion no-doc openapi] :as data} :data
+                                            middleware :middleware
+                                            interceptors :interceptors}]]
+                                (if (and data (not no-doc))
+                                  [method
+                                   (meta-merge
+                                    #_base-openapi-spec
```
ikitommi, line 100: why is this commented out?

test/cljc/reitit/ring_coercion_test.cljc
```diff
@@ -582,6 +584,79 @@
         (is (= {:status 200, :body {:total "FOO: this, BAR: that"}} (call m/schema custom-meta-merge-checking-schema)))
         (is (= {:status 200, :body {:total "FOO: this, BAR: that"}} (call identity custom-meta-merge-checking-parameters)))))))
 
+(deftest per-content-type-test
+  (doseq [[coercion json-request edn-request default-request json-response edn-response default-response]
+          [[#'malli/coercion
```
ikitommi, line 589: why Vars here?

modules/reitit-spec/src/reitit/coercion/spec.cljc
```diff
@@ -105,6 +107,40 @@
                               (if (:schema $)
                                 (update $ :schema #(coercion/-compile-model this % nil))
                                 $))]))})))
+        :openapi (openapi/openapi-spec
```
ikitommi, line 110: could be extracted into a standalone function

modules/reitit-schema/src/reitit/coercion/schema.cljc
```diff
@@ -67,6 +69,37 @@
                               (if (:schema $)
                                 (update $ :schema #(coercion/-compile-model this % nil))
                                 $))]))})))
+        :openapi (merge
```
ikitommi, line 72: could be extracted into a standalone function

ikitommi (approved):
Looks Good! pushed a small commit to add openapi module to the build.

examples/http-swagger/src/example/server.clj
```diff
@@ -42,11 +43,25 @@
...
         {:get {:no-doc true
                :swagger {:info {:title "my-api"
-                                :description "with reitit-http"}}
+                                :description "with reitit-http"}
+                         ;; used in /secure APIs below
+                         :securityDefinitions {"auth" {:type :apiKey
+                                                       :in :header
+                                                       :name "Example-Api-Key"}}}
                :handler (swagger/create-swagger-handler)}}]
+       ["/openapi.json"
```
ikitommi, line 52: we could change the default into swagger-ui handler to read the `openapi.json` over `swagger.json` if that exists?

ikitommi:
more examples could be ported to use (just) the openapi in later documentation PRs?

RickMoynihan:
I've been having a look at this PR too... Firstly thanks for putting it together, I'm very excited to see this is happening!

It seems that in order to get the swagger UI to work with the `http-swagger` example in the repo you also need to bump the swagger-ui dependency to be `[metosin/ring-swagger-ui "5.0.0-alpha.0"]` so it works with openapi `3.1.0`, it looks like @opqdonut has already done the bulk of the work by updating this dep upstream, but I think it may want updated here too.

opqdonut:
@RickMoynihan Yeah, I'm depending on the alpha swagger-ui in the example right here:
https://github.com/metosin/reitit/pull/588/files#diff-110539d4d838e0f807261c996685cb4c5e32ee025a097f1c5541b816a6ebd6aaR7

RickMoynihan:
ahh so you are!  Not sure how I missed that!  Apologies.

test/cljc/reitit/ring_coercion_test.cljc
opqdonut: Slightly nicer prints from the test:
```clj
user=> (str malli/coercion)
"reitit.coercion.malli$create$reify__9137@59568a15"
user=> (str #'malli/coercion)
"#'reitit.coercion.malli/coercion"
```

modules/reitit-openapi/src/reitit/openapi.cljc
opqdonut: no idea, left over from the previous PR #563. deleted.

modules/reitit-core/src/reitit/coercion.cljc
opqdonut: done

opqdonut:
Fixed a couple of Tommi's comments. I intend to merge this latest tomorrow.

RickMoynihan:
@opqdonut: I've noticed a small problem with the `http-swagger` example which I thought I'd report here.

If you change the `:url` for the swagger UI by passing `:url "/openapi.json"` to `swagger-ui/create-swagger-ui-handler`, run the app and try to hit the route `/files/upload` you get a 400 bad request with the following spec error:

```
{
  "spec": "(spec-tools.core/spec {:spec (clojure.spec.alpha/keys :req-un [:spec$28049/file]), :type :map, :leaf? false})",
  "problems": [
    {
      "path": [],
      "pred": "(clojure.core/fn [%] (clojure.core/contains? % :file))",
      "val": {},
      "via": [],
      "in": []
    }
  ],
  "type": "reitit.coercion/request-coercion",
  "coercion": "spec",
  "value": {},
  "in": [
    "request",
    "multipart-params"
  ]
}
```

Additionally, the swagger UI doesn't render the nice upload button that you get if you set the url to `/swagger.json`; I can imagine this later one is an upstream issue with the UI, but the former looks to be something to do with the multipart interceptor.

kjothen:
I think the example comment is wrong...it should read `create-openapi-handler`, not `openapi-spec-handler`.

``` clojure
       [\"/openapi.json\"
        {:get {:no-doc true
               :openapi {:info {:title \"my-api\"}}
               :handler reitit.openapi/openapi-spec-handler}}]
```

ikitommi:
guide on OAS3 multipart here: https://swagger.io/docs/specification/describing-request-body/multipart-requests/

ikitommi:
small commit to configure the example swagger-ui to present both swagger & openapi definitions.

Screenshot 2023-03-14 at 21 03 06

opqdonut:
I'll merge this now since it's a useful baseline. I'll fix multipart  support in a follow-up PR.

examples/http-swagger/src/example/server.clj
```diff
@@ -52,7 +52,7 @@
        ["/openapi.json"
         {:get {:no-doc true
                :openapi {:info {:title "my-api"
-                                :description "with reitit-http"
+                                :description "openap-docs with reitit-http"
```
RickMoynihan, line 56: I know this is merged, but there's a small typo here:

Should be `openapi-docs ...`

examples/http-swagger/src/example/server.clj
```diff
@@ -142,6 +175,9 @@
       (swagger-ui/create-swagger-ui-handler
         {:path "/"
          :config {:validatorUrl nil
+         	        :urls [{:name "swagger", :url "swagger.json"}
```
Deraen, line 178: Bad indent / tabs instead of spaces here.

opqdonut: in #592

examples/http-swagger/src/example/server.clj
opqdonut: in #592

PR #590, opened 2023-03-06
Update composing_routers.md

dgb23:
I wondered how one would wrap a ring handler that can be recreated at runtime without restarting the server. @ikitommi suggested to use the recently added `reloading-ring-handler` as a starting point.

I propose a small example here that illustrates the pattern.

### DIFF
--- a/doc/advanced/composing_routers.md
+++ b/doc/advanced/composing_routers.md
@@ -405,9 +405,53 @@ All the beer-routes now match in constant time.
 |-----------------|---------|-----------------------
 | `/beers/sahti`  | 40ns    | static
 
+### Wrapping a swappable ring handler
+
+In order for a ring handler to be recomposed, we can wrap it into a handler that dereferences it on request.
+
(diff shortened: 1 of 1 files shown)


doc/advanced/composing_routers.md
```diff
@@ -405,9 +405,53 @@ All the beer-routes now match in constant time.
 |-----------------|---------|-----------------------
 | `/beers/sahti`  | 40ns    | static
 
+### Wrapping a swappable ring handler
+
+In order for a ring handler to be recomposed, we can wrap it into a handler that dereferences it on request.
+
+```clj
+(defn deref-handler [rf]
```
ikitommi, line 413: this could be just:

```clojure
(defn deref-handler [rf]
  (reitit.ring/reloading-ring-handler (fn [] @rf)))
```

, but I think your example is simpler.

ikitommi:
thanks!

PR #597, opened 2023-03-18
Update Deps

ikitommi:


ikitommi:
@Deraen is ok to depend on the latest Swagger-ui?

opqdonut:
I don't think we want to ship the 5.0.0-alpha yet. I just needed it packages to get OpenAPI 3.1 support.

ikitommi:
ok, reverting.

ikitommi:
... and fixed the hasty commit: https://github.com/metosin/reitit/commit/76b5dfa1e0cfa3331edad6419cbd4ecc781c26c1

Issue #612, opened 2023-04-20
Support for descriptions in OpenAPI query parameters


MarcGn:
Hi I'm unable to add a description to the query parameters which will show up like this on the swagger UI:

![image](https://user-images.githubusercontent.com/59601490/233409355-37f2a87b-de6d-4d30-bb00-3fe44e48fed0.png)

Using schema-tools `schema`, I can add a description like so:
```clojure
(s/defschema QuerySchema
 {:start-time (st/schema s/Num {:openapi/description "Description here"})
  :end-time (st/schema s/Num {:openapi/description "Description here"})}})
```

But this will generate an output where the description is a property under :schema  and the top level description is still `""` (so nothing shows in the UI):

```json
{
          "name" : "start-time",
          "in" : "query",
          "description" : "",
          "required" : true,
          "schema" : {
            "type" : "number",
            "format" : "double",
            "description" : "Description here"
          }
 }
```

p-himik:
The same thing happens with Malli. Pretty sure this block is the root cause: https://github.com/metosin/reitit/blob/master/modules/reitit-malli/src/reitit/coercion/malli.cljc#L151-L155

opqdonut:
fix for malli in PR #624 

schema needs fixes on in schema-tools

opqdonut:
spec seems to work currently. I've added a test for it in #624

opqdonut:
#625 fixes this for spec

PR #624, opened 2023-05-17
fix: openapi malli parameter descriptions

opqdonut:
... should come from the parameter type, not from the parent :map

for #612


opqdonut:
# Before
openapi-before

# After
openapi-after

ikitommi (approved):

PR #641, opened 2023-08-29
doc: fix dev_workflow.md code example

allentiak:
Does the code from L145 really need to be different from the one shown in L32?

If I use `(constantly...)`, I get the following exception:

`Exception: java.lang.IllegalArgumentException: No implementation of method: :match-by-path of protocol: #'reitit.core/Router found for class: clojure.core$constantly$fn__5740`


allentiak:
BTW: The "cljdoc analysis" check fails. But I haven't changed any cljdoc...

Issue #651, opened 2023-09-28
Example projects using reitit.swagger fail to compile 


vedang:
As of 497da675b92d2d50b00feaf3ff662145a04a250d (committed 1st Sept
2023), `swagger.cljc` uses the `update-vals` function, introduced in
Clojure 1.11.

Since the `project.clj` files use Clojure 1.10, this causes a compilation error
when we try to run the examples.

PR #652, opened 2023-09-28
Update Clojure version of all Swagger Examples to 1.11

vedang:
As of 497da675b92d2d50b00feaf3ff662145a04a250d (committed 1st Sept 2023), `swagger.cljc` uses the `update-vals` function, introduced in Clojure 1.11.

Due to this, all the example projects fail to run since the `project.clj` specifies Clojure version 1.10.

This commit updates the Clojure version for such examples, fixing the problem.

Fixes: #651


ikitommi:
Thanks!

PR #656, opened 2023-10-13
Support independent request and response validation in Malli coercer

stig:
The motivation for this is adding a new API to a legacy service whose data model has evolved over the years. Maybe we cannot guarantee the presence of all fields that are considered mandatory now. Rather than blowing up when attempting to return legacy entries we may want to disable response validation.

However, we do want to ensure *new* entries added to the system are valid, so we do not want to disable request validation.

Retain the ability to turn off both request and response validation with a single setting for backwards compatibility.


stig:
I take the radio silence means "not interested". I'll close this, to get it off my list of Open PRs I need to check back on every week. Feel free to re-open if you decide you're interested.

Issue #660, opened 2024-02-03
Is it possible to do multi host with reitit-routes?


matheusfrancisco:
I did found on the documentation anything about do a multiple host routes. 
I would like to do a set of routes to when I hit `api.xpto.com` it will redirect to be available some set of routes, and when I hit `api2.xpto.com` it will be available another set of routes. 

Currently, I have an application running in production with pedestal routes, so I was switching to reitit, but I miss this feature, if is something you guys would like to have or any suggestion I'm willing to open a pull request or something.
```clojure
(ns myapp.service
  (:require [io.pedestal.http.route :as route]))

(def application-routes
  (route/expand-routes
    #{{:host "example.com" :scheme :https}
      ["/user"          :get user-search-form]
      ["/user/:user-id" :get view-user        :constraints {:user-id #"[0-9]+"}]
      ,,,
      }))
```

ikitommi:
There is nothing built-in, but you can build this easily on top of reitit. e.g.

1. create a middleware that reads `:host` and `:schema` keys from route data. see parameter coercion middleware as example, e.g. either needs to needs to be set for the middleware to mount into to the route
2. mount the middleware to top-level
3. attach data to route tree, at any level, e.g.

```clojure
[["" {:host "example.com"} ...]
 ["/all-hosts"
  ["/trade" {:host "trade.com"} ...]
  ["/api"
   ["/admin" {:host "admin.com" :scheme :https} ...]
   ["/public" {:host "public.com"} ...]]]]
```

matheusfrancisco:
Thanks for answer. 
I looked on the docs and this is not there, if you want I can open a pr adding documentation to multi-host with pedestal, or I can add a small example inside the sample folder

matheusfrancisco:
something like this 
```clojure
(defn host-scheme-middleware
  "Middleware to check host and scheme against route data."
  [handler]
  (fn [request]
    (let [route-data (:reitit.core/match request)
          route-meta (and route-data (:data route-data))
          request-host (:headers request "host")
          request-scheme (if (-> request :request-method (= "https")) :https :http)
          route-host (:host route-meta)
          route-scheme (:scheme route-meta)]
      (if (and (or (not route-host) (= request-host route-host))
               (or (not route-scheme) (= request-scheme route-scheme)))
        (handler request)
        (response/bad-request "Host or scheme does not match")))
```

matheusfrancisco:
[...]

PR #661, opened 2024-02-04
E.g: multihost with pedestal.

matheusfrancisco:
Currently, I run into switch my application to reitit, but we were using multi-host with pedestal, and then I open this https://github.com/metosin/reitit/issues/660 for people who are using this feature from pedestal-routes and would like to switch to reitit


matheusfrancisco:
missing some commit

PR #666, opened 2024-02-24
Update ring-swagger-ui for support openapi 3.1.0 version

velios:
The need for such a change is that reitit-openapi module creates openapi.js of 3.1.0 version, but current reitit-swagger-ui can't draw it, because dependency through root to ring-swagger-ui . ring-swagger-ui > 5.x.x version can draw openapi.js 3.1.0 version correctly

I add pic in #[1011](https://github.com/metosin/malli/issues/1011) issue, which I created before I figured out the problem


velios:
Anybody here?

opqdonut:
Sorry for the delay! You're right, we should upgrade ring-swagger-ui. The examples/ all use 5.9.0, but for some reason the main project.clj still depended on an older version.

PR #674, opened 2024-04-19
Bump deps on examples

opqdonut:


opqdonut:
TODO:
frontend-controllers
frontend-links
frontend-malli
frontend-prompt
frontend-re-frame

mvarela:
Frontend deps bumped, and code updated to work with current reagent version.

PR #681, opened 2024-05-11
Add reitit-openapi to readme, clarify group-ids

alexhall:
This may help prevent confusion for users who want to individually install modules.


opqdonut:
thanks!

PR #685, opened 2024-06-21
Add arities 1 and 2 to rf/match->path

PEZ:
To adhere to the doc string's info about parameter 2 and 3 being optional.


ikitommi:
Thanks!

PR #693, opened 2024-08-25
Speed up routes and inline it in code ring handler

### LINKED ISSUE #692: Async arity ring handler is slower than sync arity
Possible culprit:
https://github.com/metosin/reitit/blob/master/modules/reitit-ring/src/reitit/ring.cljc#L366
Specifically, the problem is with calling `routes` dynamically
alternative solution, build a unit of composition and use a transducer
```clojure
(defn comp-handlers
  "Create a ring handler by combining several handlers into one."
  ([handler]
   handler)
  ([handler1 handler2]
   (let [single-arity (fn [request]
                        (or (handler1 request) (handler2 request)))
         multi-arity (fn [request respond raise]
                       (handler1 request (fn [response]
                                           (if response
                                             (respond response)
                                             (handler2 request respond raise))) raise))]
     (fn
       ([request]
        (single-arity request))
       ([request respond raise]
        (multi-arity request respond raise))))))
```

bsless:
Fixes #692

### DIFF
--- a/modules/reitit-ring/src/reitit/ring.cljc
+++ b/modules/reitit-ring/src/reitit/ring.cljc
@@ -363,7 +363,10 @@
                   result (:result match)
                   handler (-> result method :handler (or default-handler))
                   request (enrich-request request path-params match router)]
-              ((routes handler default-handler) request respond raise))
+              (handler request (fn [response]
+                                 (if response
+                                   (respond response)
+                                   (default-handler request respond raise))) raise))
             (default-handler (enrich-default-request request router) respond raise))
           nil)))
       {::r/router router}))))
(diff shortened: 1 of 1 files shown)


modules/reitit-ring/src/reitit/ring.cljc
```diff
@@ -133,21 +133,38 @@
...
+                                           (if response
+                                             (respond response)
+                                             (handler2 request respond raise))) raise))]
+     (fn
+       ([request]
+        (single-arity request))
+       ([request respond raise]
+        (multi-arity request respond raise))))))
+
 (defn routes
```
ikitommi, line 155: I think this could be simplified a bit.

bsless: In what way?

ikitommi: `comp-handlers` has 2 arities, `routes` 4. `routes` is not needed at request-time (thanks for the fix!) so the creation speed is not critical. dropped extra arities here: https://github.com/metosin/reitit/pull/696

PR #701, opened 2024-09-13
Add dispatch for every implementation of IPersistentMap

### LINKED ISSUE #700: Header coercion no longer works using aleph with 0.7.2
Might be a regression introduced by #506 ?

aleph headers are potemkin magical maps where keys can be accessed both as strings and as keywords.

With a reitit route that has `:parameters {:header {:authorization (,,, spec-tools spec for Bearer ,,,)}}` and `:coercion spec-coercion/coercion`, with 0.7.1 `coercion/coerce-request-middleware` works as expected and valid requests pass.

However with 0.7.2 this no longer works failing with "Request coercion failed". In 0.7.2 if you add a middleware like below it once again works as expected, so I suspected it might be some optimizations in #506 clashing with aleph's (potemkin's) custom map types.
```clj
(fn [handler]
  (fn [request]
    (handler (update request :headers #(into {} %)))))
```

bsless:
Closes #700

### DIFF
--- a/modules/reitit-core/src/reitit/walk.clj
+++ b/modules/reitit-core/src/reitit/walk.clj
@@ -9,7 +9,7 @@
   [m k v]
   (assoc! m (if (string? k) (keyword k) (-keywordize k)) (-keywordize v)))
 
-(defn- -keywordize-map
+(defn- -keywordize-kvreducible
   [m]
   (persistent! (reduce-kv keywordize-kv (transient (empty m)) m)))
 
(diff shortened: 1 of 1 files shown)


modules/reitit-core/src/reitit/walk.clj
```diff
@@ -29,6 +29,14 @@
...
               clojure.lang.PersistentHashMap
               clojure.lang.PersistentTreeMap]]
+  (extend type IKeywordize {:-keywordize -keywordize-kvreducible}))
+
+(defn- -keywordize-map
+  [m]
+  (let [f (fn [[k v]] (if (string? k) [(keyword k) v] [k v]))]
+    (into {} (map f) m)))
+
+(doseq [type [clojure.lang.IPersistentMap]]
```
ikitommi, line 39: Why doseq here?

bsless: Mostly because I copied the other implementation.
Makes it easier to extend to another type in the future

PR #703, opened 2024-10-10
Add OpenAPI :requestBody for :form request schema

mpenttila:
[Ring Coercion](https://github.com/metosin/reitit/blob/master/doc/ring/coercion.md) functionality allows defining a request schema with `{:parameters {:form some-schema}}`. This resulted in that schema being transformed to OpenAPI parameters, which was supported in OAS 2 (Swagger), [but is not supported in OAS 3](https://swagger.io/docs/specification/v3_0/describing-request-body/describing-request-body/#differences-from-openapi-20). Now they must be defined as a `requestBody` schema like any other kind of body. 

`:form` parameter coercion also implies that the request content type must be `application/x-www-form-urlencoded`.

### DIFF
--- a/modules/reitit-openapi/src/reitit/openapi.clj
+++ b/modules/reitit-openapi/src/reitit/openapi.clj
@@ -76,10 +76,12 @@
 (defn- openapi-path [path opts]
   (-> path (trie/normalize opts) (str/replace #"\{\*" "{")))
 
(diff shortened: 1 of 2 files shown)


opqdonut (approved):
Thanks! This is great. Sorry for the long response time :(

PR #712, opened 2024-12-07
Make exception middleware return proper Ring responses

### LINKED ISSUE #708: Handlers in `reitit.ring.middleware.exception` return incorrect Ring responses
According to the Ring spec, a response is something that has a body, a status, and a headers map. In other words, something for which `ring.util.response/response?` returns true.

But some handlers in `reitit.ring.middleware.exception` return response maps without `:headers`.

Same thing with `reitit.ring.coercion/handle-coercion-exception`.

p-himik:
Fixes #708


ikitommi:
Thanks!

PR #714, opened 2025-01-17
Place the openapi deprecation tag directly under parameters

mokshasoft:
When creating an openapi handler via reitit.openapi/create-openapi-handler the parameter deprecation tags are put under the schema tag, not directly under the parameters tag.

https://swagger.io/docs/specification/v3_0/describing-parameters/#deprecated-parameters

The openapi.json generated without this fix:
```json
          {
            "in": "query",
            "name": "include-legacy-information",
            "required": false,
            "schema": {
              "deprecated": true
            },
          },
```

The openapi.json generated with this fix:
```json
          {
            "in": "query",
            "name": "include-legacy-information",
            "required": false,
            "schema": {},
            "deprecated": true
          },
```

I was able to run these tests, that passed:
./scripts/test.sh clj 
but not these, that failed for an unrelated reason:
./scripts/test.sh cljs

### DIFF
--- a/doc/ring/openapi.md
+++ b/doc/ring/openapi.md
@@ -139,6 +139,7 @@ Malli:
            [:x
             {:title "X parameter"
              :description "Description for X parameter"
+             :json-schema/deprecated true
              :json-schema/default 42}
             int?]
            [:y int?]]}}}]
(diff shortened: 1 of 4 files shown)


opqdonut:
Thanks for the fix! I'm sorry I didn't notice this earlier. I'll try to get it in ASAP.

opqdonut:
thanks once more! will be out in the next release!

mokshasoft:
You're welcome!

PR #716, opened 2025-01-21
Use coercion to encode query-string values in match->path

Deraen:
This is the initial version of extending Reitit code and the frontend module to encode query-string parameter values using the Malli schema and string transformer.

Coercion already allowed parsing query-string strings using Malli and even per schema property using the `:decode/string` option. Implementing this also for URL generation (href, push/replace-state, etc.) makes sense because Malli also supports value encoding using the schemas.

Changes:

- Coercion protocol needs to be extended with a method to get coercer fn for this use case
Coercion implementation for Malli. Spec and Schema are a no-op at this point. We need to consider whether they can also support this case.
- The match->path function now gets a coercer for the match (creates it again for every call!!) and uses the coercer to encode query-string map values first before using the existing query-params url-encode function
    - TODO: Does forced url-encoding for the values make sense?
    - What if we had a separate URL-encoded transformer for Malli coercion? Current string-transformer options might differ from query-string use case.

TODO

- [x] Frontend module tests
- [x] Figure out where to place the match->path version with coercion, likely not reitit.core
- [x] Use -compile-model correct to merge schemas etc., before use in match->path coercion
- [x] ~Consider if query-coercer can be "precompiled" in reitit.frontend/router, so the coercer and transformers don't need to be recreated each time~
- [x] Check if we should avoid the malli/-coercer implementation, which does decode -> validate -> encode. For query strings, only encode might be better. Query string parameters are likely already decoded, so with this version, decode functions need to check if the value is encoded or not.

Separate or further work

- [x] Frontend set-query-params and set-query currently don't create a Match from the current path, so schema and coercion options aren't available. It should be possible to extend these functions to get a match first, so schema and coercion options are available. (At least for reitit.frontend.easy.)

### DIFF
--- a/doc/frontend/coercion.md
+++ b/doc/frontend/coercion.md
@@ -1,0 +1,59 @@
+# Frontend coercion
+
+The Reitit frontend leverages [coercion](../coercion/coercion.md) for path,
+query, and fragment parameters. The coercion uses the input schema defined
+in the match data under `:parameters`.
+
+## Behavior of Coercion
+
+1. **Route Matching**
+   When matching a route from a path, the resulting match will include the
+   coerced values (if coercion is enabled) under `:parameters`. If coercion is
+   disabled, the parsed string values are stored in the same location.
+   The original un-coerced values are always available under `:path-params`,
+   `:query-params`, and `:fragment` (a single string).
+
+2. **Creating Links and Navigating**
+   When generating a URL (`href`) or navigating (`push-state`, `replace-state`, `navigate`)
+   to a route, coercion can be
+   used to encode query-parameter values into strings. This happens before
+   Reitit performs basic URL encoding on the values. This feature is
+   especially useful for handling the encoding of specific types, such as
+   keywords or dates, into strings.
+
+3. **Updating current query parameters**
+  When using `set-query` to modify current query parameters, Reitit frontend
+  first tries to find a match for the current path so the match can be used to
+  first decode query parameters and then to encode them. If the current path
+  doesn't match the routing tree, `set-query` keeps all the query parameter
+  values as strings.
+
+## Notes
+
+- **Value Encoding Support**: Only Malli supports value encoding.
+- **Limitations**: Path parameters and fragment values are not encoded using
+  the match schema.
+
+## Example
+
+```cljs
+(def router (r/router ["/"
+                       ["" ::frontpage]
+                       ["bar"
+                        {:name ::bar
+                         :coercion rcm/coercion
+                         :parameters {:query [:map
+                                              [:q {:optional true}
+                                               [:keyword
+                                                {:decode/string (fn [s] (keyword (subs s 2)))
+                                                 :encode/string (fn [k] (str "__" (name k)))}]]]}}]]))
+
+(rfe/href ::bar {} {:q :hello})
+;; Result "/bar?q=__hello", the :q value is first encoded
+
+(rfe/push-state ::bar {} {:q :world})
+;; Result "/bar?q=__world"
+;; The current match will contain both the original value and parsed & decoded parameters:
+;; {:query-params {:q "__world"} 
+;;  :parameters {:query {:q :world}}}
+```
--- a/modules/reitit-core/src/reitit/core.cljc
+++ b/modules/reitit-core/src/reitit/core.cljc
@@ -78,14 +78,15 @@
                          ;; on the coercion?
                          ;; NOTE: Re-creates coercer on every call, could this be pre-compiled somewhere
                          ;; or memoized? Does it matter much?
+                         ;; TODO: query-coercer could be compiled in reitit.frontend/router, same as request coercers.
                          (str "?" (let [coercion (-> match :data :coercion)
                                         schema (when coercion
                                                  (coercion/-compile-model coercion (-> match :data :parameters :query) nil))
                                         coercer (when (and schema coercion)
                                                   (coercion/-query-string-coercer coercion schema))
-                                        query-params (or (when coercer
-                                                           (coercer query-params :default))
-                                                         query-params)]
+                                        query-params (if coercer
+                                                       (coercer query-params :default)
+                                                       query-params)]
                                     ;; Default encoding for values will handle values that aren't encoded using coercer
                                     (impl/query-string query-params)))))))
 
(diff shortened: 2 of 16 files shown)


modules/reitit-schema/src/reitit/coercion/schema.cljc
```diff
@@ -101,6 +101,9 @@
             value))))
     (-response-coercer [this schema]
       (if (coerce-response? schema)
-        (coercion/-request-coercer this :response schema)))))
+        (coercion/-request-coercer this :response schema)))
+    (-query-string-coercer [this schema]
+      ;; TODO: Can this be implemented?
```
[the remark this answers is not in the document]
Deraen: It would be really spammy; with the current implementation, this is called each time a path with a query string is created. Though we could trigger warning just once, delay or something.

modules/reitit-core/src/reitit/core.cljc
```diff
@@ -1,7 +1,9 @@
 (ns reitit.core
   (:require [reitit.exception :as exception]
             [reitit.impl :as impl]
-            [reitit.trie :as trie]))
+            [reitit.trie :as trie]
+            ;; FIXME: Should avoid coercion require here?
+            [reitit.coercion :as coercion]))
```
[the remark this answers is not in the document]
ikitommi: discussed with Juho, let's keep the core small. coercion is an extra application on top of it.

doc/frontend/coercion.md
```diff
@@ -0,0 +1,59 @@
...
+
+(rfe/href ::bar {} {:q :hello})
+;; Result "/bar?q=__hello", the :q value is first encoded
+
+(rfe/push-state ::bar {} {:q :world})
+;; Result "/bar?q=__world"
+;; The current match will contain both the original value and parsed & decoded parameters:
+;; {:query-params {:q "__world"} 
+;;  :parameters {:query {:q :world}}}
+```
```
opqdonut, line 59: good examples!

modules/reitit-frontend/src/reitit/frontend/easy.cljs
```diff
@@ -48,9 +48,10 @@
   The URL is formatted using Reitit frontend history handler, so using it with
   anchor element href will correctly trigger route change event.
 
-  Note: currently collections in query-parameters are encoded as field-value
-  pairs separated by &, i.e. \"?a=1&a=2\", if you want to encode them
-  differently, convert the collections to strings first."
+  By default currently collections in query parameters are encoded as field-value
```
opqdonut, line 51: I think "By default, collections ..." would read a bit better

opqdonut (approved):
LGTM!

modules/reitit-malli/src/reitit/coercion/malli.cljc
```diff
@@ -76,6 +76,20 @@
...
                 value))))))))
 
+(defn- -query-string-coercer
+  "Create coercer for query-parameters, always allows extra params and does
+  encoding using string-transformer."
+  [schema transfomers options]
+  (let [;; Always allow extra paramaters on query-parameters encoding
+        open-schema (mu/open-schema schema)
+        ;; Do not remove extra keys
+        string-transformer (-transformer string-transformer-provider (assoc options :strip-extra-keys false))
```
lucassousaf, line 86: Hello! We've been testing these changes and we have a problem when using custom transformers. We think the problem might be this line. This line is ignoring the `transfomers` (there is a typo too) parameter and hard-coding `string-transfomer-provider` instead of using the `transformers` defined in the Reitit router configuration. Maybe it should be something like this?:
```clj
(defn- -query-string-coercer
  "Create coercer for query-parameters, always allows extra params and does
  encoding using string-transformer."
  [schema transfomers options]
  (let [;; Always allow extra paramaters on query-parameters encoding
        open-schema (mu/open-schema schema)
        ;; Do not remove extra keys
        ;; BEFORE: (-transformer (get-in transfomers [:string :default]) (assoc options :strip-extra-keys false))
        string-transformer (get-in transfomers [:string :default])
        encoder (m/encoder open-schema (assoc options :strip-extra-keys true) string-transformer)]
    (fn [value format]
      (if encoder
        (encoder value)
        value)))) 
 ```
 Is there any reason for hard-coding `string-transformer-provider`?

lucassousaf, line 86: Hello! We've been testing these changes and we have a problem when using custom transformers. We think the problem might be this line. This line is ignoring the `transfomers` (there is a typo too) parameter and hard-coding `string-transfomer-provider` instead of using the `transformers` defined in the Reitit router configuration. Maybe it should be something like this?:
```clj
(defn- -query-string-coercer
  "Create coercer for query-parameters, always allows extra params and does
  encoding using string-transformer."
  [schema transfomers options]
  (let [;; Always allow extra paramaters on query-parameters encoding
        open-schema (mu/open-schema schema)
        ;; Do not remove extra keys
        ;; BEFORE: (-transformer (get-in transfomers [:string :default]) (assoc options :strip-extra-keys false))
        string-transformer (get-in transfomers [:string :default])
        encoder (m/encoder open-schema (assoc options :strip-extra-keys true) string-transformer)]
    (fn [value format]
      (if encoder
        (encoder value)
        value)))) 
 ```
 Is there any reason for hard-coding `string-transformer-provider`?

Deraen: The reason is that `create` fn processes `transformers` map and applies the options to TransformerProviders, then the transformer we see after that step is one with `mt/strip-extra-keys-transformer` combined with the string-transformer.

- We can probably use the transformers value before options are applied for this code
- the options are just applied to the default TransformationProvider values, when you provide a concrete transformer instance yourself, this doesn't apply
- while `m/encoder` takes options map, `:strip-extra-keys` isn't one of the options that would be used there (AFAIK)

I'll see what I can do.

lucassousaf: I see, you're right. The encoder doesn't take the `:strip-extra-keys` option.

Maybe one solution could be to have a specific query string transformer provider that doesn't include the `:strip-extra-keys` transformer and include it by default in the Reitit router configuration.

Deraen: @lucassousaf Check the latest commit, :string transformer will be used now. I did consider adding a new transfomers option for query-string encoding, but I think decoding and encoding should use the same transformer so using :string does seem good idea now.

lucassousaf: Sounds good! At first glance it seems to work as expected. If we find anything (I don't think so, looks good) we can open an issue. Thanks a lot for your work @Deraen !

Issue #540, opened 2022-03-18
Create simplified router for frontend use


Deraen:
Frontend routing tree is often relatively small, and instead of thousands (or more) request per second, we only need to occasionally find the match when user navigates in the app. So performance is not as important.

Currently reitit router always uses the trie implementation. This is performance optimized code, and the resulting JS code is quite large.

![image](https://user-images.githubusercontent.com/648087/159033906-e60dc1e2-456b-47d6-af36-15d8f1001dff.png)

PR #723, opened 2025-02-25
Add regex based router

### LINKED ISSUE #722: Routing with regex based constraints
I have an existing routing structure I'm trying to migrate from Bidi. It uses regular expressions quite heavily. Bidi supports these quite happily: https://github.com/juxt/bidi#regular-expressions. However, I'm not sure how to translate these into Reitit routes.

Here's one of our simpler examples:

```clj
(let [rtr (r/router ["/"
                     ["" ::home]
                     ["inbox" ::inbox]
                     ["teams" ::teams]
                     ["teams/:team-id-b58/members" {:name ::->members
                                                    :parameters {:path {:team-id-b58 (partial re-matches #"[a-z]")}}}]])]
  [(r/match-by-path rtr "/")
   (r/match-by-path rtr "/inbox")
   (r/match-by-path rtr "/teams/123/members")
   (r/match-by-path rtr "/teams/abcdefg/members")])
```

<details>

<summary>output</summary>

```clj
[#reitit.core.Match{:template "/", :data {:name :user/home}, :result nil, :path-params {}, :path "/"}
 #reitit.core.Match{:template "/inbox", :data {:name :user/inbox}, :result nil, :path-params {}, :path "/inbox"}
 #reitit.core.Match{:template "/teams/:team-id-b58/members",
                    :data {:name :user/->members,
                           :parameters {:path [{:team-id-b58 #object[clojure.core$partial$fn__5927
                                                                     0x4bedf1bc
                                                                     "clojure.core$partial$fn__5927@4bedf1bc"]}]}},
                    :result nil,
                    :path-params {:team-id-b58 "123"},
                    :path "/teams/123/members"}
 #reitit.core.Match{:template "/teams/:team-id-b58/members",
                    :data {:name :user/->members,
                           :parameters {:path [{:team-id-b58 #object[clojure.core$partial$fn__5927
                                                                     0x4bedf1bc
                                                                     "clojure.core$partial$fn__5927@4bedf1bc"]}]}},
                    :result nil,
                    :path-params {:team-id-b58 "abcdefg"},
                    :path "/teams/abcdefg/members"}]

```

</details>

Here there are no conflicts, but Reitit is matching both of the routes. If I try to load `/teams/123/members` in our frontend app, it crashes with an exception once request coercion runs:

```clj
:app.boot/init-router event error: Request coercion failed 
:type :reitit.coercion/request-coercion]
[:coercion ]
[:value {:team-id-b58 "123"}]
[:in [:request :path-params]]
```

Ideally I'd like those `/teams/123/members` to not be matched at all so I can serve a default "not found" route.

Then we have a more challenging example:

```clj
(let [rtr (r/router ["/"
                     ["" ::home]
                     [":item-id" {:name ::item
                                  :parameters {:path {:item-id (partial re-matches #"[a-z]{16,20}")}}}]
                     ["inbox" ::inbox {:conflicting true}]
                     ["teams" ::teams {:conflicting true}]
                     ["teams/:team-id-b58/members" {:name ::->members
                                                    :parameters {:path {:team-id-b58 (partial re-matches #"[a-z]")}}}]
                     ])]
  [(r/match-by-path rtr "/")
   (r/match-by-path rtr "/inbox")
   (r/match-by-path rtr "/itemwithlongname")
   (r/match-by-path rtr "/teams/123/members")
   (r/match-by-path rtr "/teams/abcdefg/members")])
```

<details>

<summary>output</summary>

```clj
[#reitit.core.Match{:template "/", :data {:name :user/home}, :result nil, :path-params {}, :path "/"}
 #reitit.core.Match{:template "/:item-id",
                    :data {:name :user/item,
                           :parameters {:path [{:item-id #object[clojure.core$partial$fn__5927
                                                                 0x4f1cba3b
                                                                 "clojure.core$partial$fn__5927@4f1cba3b"]}]}},
                    :result nil,
                    :path-params {:item-id "inbox"},
                    :path "/inbox"}
 #reitit.core.Match{:template "/:item-id",
                    :data {:name :user/item,
                           :parameters {:path [{:item-id #object[clojure.core$partial$fn__5927
                                                                 0x4f1cba3b
                                                                 "clojure.core$partial$fn__5927@4f1cba3b"]}]}},
                    :result nil,
                    :path-params {:item-id "itemwithlongname"},
                    :path "/itemwithlongname"}
 #reitit.core.Match{:template "/teams/:team-id-b58/members",
                    :data {:name :user/->members,
                           :parameters {:path [{:team-id-b58 #object[clojure.core$partial$fn__5927
                                                                     0x5a45ffe6
                                                                     "clojure.core$partial$fn__5927@5a45ffe6"]}]}},
                    :result nil,
                    :path-params {:team-id-b58 "123"},
                    :path "/teams/123/members"}
 #reitit.core.Match{:template "/teams/:team-id-b58/members",
                    :data {:name :user/->members,
                           :parameters {:path [{:team-id-b58 #object[clojure.core$partial$fn__5927
                                                                     0x5a45ffe6
                                                                     "clojure.core$partial$fn__5927@5a45ffe6"]}]}},
                    :result nil,
                    :path-params {:team-id-b58 "abcdefg"},
                    :path "/teams/abcdefg/members"}]

```

</details>

We have a root "item-id" route which is matched by a regular expression. This lives at the root of the app, so we can't prefix it with `/item/:item-id` and put it under another route prefix. Is there a way to achieve this route with Reitit? I understand it might be a bit slower because it would need to evaluate a regex or similar on the fly, but it is the routing structure we need to live with.

Similar to, but different from #721.

danielcompton:
Reitit is very fast, but cannot support all routing structures that other routers like Bidi support, particular regex-based routes.

This PR is a proof of concept of one way that a regex based router could work with Reitit.

A more complete version of this PR would compile only the regex based routes to this router, and let the rest of the routes be handled by faster Reitit routers. You could also compile the regex based router to match on a path segment at a time, pruning routes before needing to evaluate a regular expression.

Before I go too much further, I wanted to get some feedback on the syntax for route definition and the overall approach. What do you think about this idea, and the way that path parameters are defined as regex's?

This unoptimized router is about 8-20x slower than regular Reitit in some informal testing.

Fixes https://github.com/metosin/reitit/issues/722

### DIFF
--- a/modules/reitit-core/src/reitit/regex.cljc
+++ b/modules/reitit-core/src/reitit/regex.cljc
@@ -1,0 +1,121 @@
+(ns reitit.regex
+  (:require [clojure.set :as set]
+            [clojure.string :as str]
+            [reitit.core :as r]))
+
+(defn compile-regex-route
+  "Given a route vector [path route-data], returns a map with:
+   - :pattern: a compiled regex pattern built from the path segments,
+   - :group-keys: vector of parameter keys in order,
+   - :route-data: the provided route data,
+   - :original-segments: original path segments for path generation,
+   - :template: the original path template for Match objects."
+  [[path route-data]]
+  (let [;; Normalize route-data to ensure it's a map with :name
+        route-data (if (keyword? route-data)
+                     {:name route-data}
+                     route-data)
+
+        ;; Store the original path template for Match objects
+        template (if (str/starts-with? path "/")
+                   path
+                   (str "/" path))
+
+        ;; Handle paths with or without leading slashes
+        normalized-path (cond-> path
+                                (str/starts-with? path "/") (subs 1))
+
+        ;; Split into segments, handling empty paths
+        segments (if (empty? normalized-path)
+                   []
+                   (str/split normalized-path #"/"))
+
+        ;; Store original segments for path generation
+        original-segments segments
+
+        compiled-segments
+        (map (fn [seg]
+               (if (str/starts-with? seg ":")
+                 (let [param-key   (keyword (subs seg 1))
+                       param-regex (get-in route-data [:parameters :path param-key])]
+                   (if (and param-regex (instance? java.util.regex.Pattern param-regex))
+                     (str "(" (.pattern ^java.util.regex.Pattern param-regex) ")")
+                     ;; Fallback: match any non-slash characters.
+                     "([^/]+)"))
+                 (java.util.regex.Pattern/quote seg)))
+             segments)
+
+        ;; Create the pattern string, handling special case for root path
+        pattern-str (if (empty? segments)
+                      "^/?$"  ;; Match root path with optional trailing slash
+                      (str "^/" (str/join "/" compiled-segments) "$"))
+
+        group-keys (->> segments
+                        (filter #(str/starts-with? % ":"))
+                        (map #(keyword (subs % 1)))
+                        (vec))]
+
+    {:pattern    (re-pattern pattern-str)
+     :group-keys group-keys
+     :route-data route-data
+     :original-segments original-segments
+     :template template}))
+
+(defn- generate-path
+  "Generate a path from a route and path parameters."
+  [route path-params]
+  (if (empty? (:original-segments route))
+    "/"
+    (str "/" (str/join "/"
+                       (map (fn [segment]
+                              (if (str/starts-with? segment ":")
+                                (let [param-key (keyword (subs segment 1))]
+                                  (get path-params param-key ""))
+                                segment))
+                            (:original-segments route))))))
+
+(defrecord RegexRouter [compiled-routes]
+  r/Router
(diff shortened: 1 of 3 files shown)


Deraen:
This could be also interesting for frontend use, to optimize the artifact size (#540) because the Trie implementation generates quite a lot of JS code and the performance doesn't matter as much for FE routing.

I'll try to review this at some point.

mthl:
Hello

```clj
[["articles/:year/:month/:slug"
  {:name :user/article,
   :parameters {:path {:year #"\d{4}"
                       :month #"\d{2}"
                       :slug #"[a-z0-9\-]+"}}}]]
```

Given that the regexp guard is defined in `[:parameters :path]` which is used by coercion and openapi, it seems that the regexp router is only compatible with schema and not with malli or clojure.spec which is not ideal.

PR #725, opened 2025-03-21
Add option to allow serving index files without redirect

### LINKED ISSUE #678: Resource handler: serve index file without a redirect
I am using ring's resource handler (`ring/create-resource-handler`) to serve a Vite application. Vite builds the application as `index.html` and other files in the same distribution directory. I rely on the resource handler's support for index files to serve this application. However, the resource handler's index files support doesn't work quite like I expect or would like.

When I request `/`, this handler responds with an HTTP 302 redirect to `/index.html`. This results in an undesired customer-visible URL. I want `index.html` to be an implementation detail, and for the resource handler to serve the content of `index.html` in response to `GET /` without a redirect. This would be consistent with how Apache httpd's `mod_dir` works; [see default for `DirectoryIndexRedirect` directive.](https://httpd.apache.org/docs/2.4/mod/mod_dir.html#directoryindexredirect)

Can you suggest how I can achieve the desired behavior of serving `index.html` transparently on `GET /`?

Some options I see:

1. Create a custom route for just `/index.html` that serves my file directly, with the existing resource handler as a fallback for other routes. I imagine then I'd need to disable conflict resolution, which doesn't sound good.
2. Add a configuration parameter to the ring resource handler to support the equivalent of apache httpd's `DirectoryIndexRedirect` options. Do you think this would be a good idea?

agorgl:
Adds an `:index-redirect?` to `reitit.ring/create-resource-handler` and `reitit.ring/create-file-handler` functions that allow serving matching index files without redirecting to them.

Fixes #678

### DIFF
--- a/modules/reitit-ring/src/reitit/ring.cljc
+++ b/modules/reitit-ring/src/reitit/ring.cljc
@@ -227,18 +227,20 @@
    ;; TODO: ring.middleware.head/wrap-head
    ;; TODO: handle etags
    (defn -create-file-or-resource-handler
-     [response-fn {:keys [parameter root path loader allow-symlinks? index-files paths not-found-handler]
+     [response-fn {:keys [parameter root path loader allow-symlinks? index-files index-redirect? paths not-found-handler]
                    :or {parameter (keyword "")
                         root "public"
                         index-files ["index.html"]
-                        paths (constantly nil)
-                        not-found-handler (if path
-                                            (constantly nil)
-                                            (constantly {:status 404, :body "", :headers {}}))}}]
+                        index-redirect? true
+                        paths (constantly nil)}}]
      (let [options {:root root
                     :loader loader
                     :index-files? false
                     :allow-symlinks? allow-symlinks?}
+           not-found-handler (or not-found-handler
+                                 (if path
+                                   (constantly nil)
+                                   (constantly {:status 404, :body "", :headers {}})))
            path-size (count path)
            create (fn [handler]
                     (fn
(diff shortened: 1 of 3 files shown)


opqdonut:
ignore the cljs test failure, it was only a deprecated github action. fixed by #726

modules/reitit-ring/src/reitit/ring.cljc
```diff
@@ -281,6 +285,7 @@
      | :path              | path to mount the handler to. Required when mounted outside of a router, does not work inside a router.
      | :loader            | optional class loader to resolve the resources
      | :index-files       | optional vector of index-files to look in a resource directory, defaults to `[\"index.html\"]`
+     | :index-redirect?   | optional boolean that indicates if to redirect on matched index file or serve it from the matched path
```
opqdonut, line 288:
```suggestion
     | :index-redirect?   | optional boolean: if true (default), redirect to index file, if false serve it directly
```

modules/reitit-ring/src/reitit/ring.cljc
```diff
@@ -298,6 +303,7 @@
...
```
opqdonut, line 306:
```suggestion
     | :index-redirect?   | optional boolean: if true (default), redirect to index file, if false serve it directly
```

modules/reitit-ring/src/reitit/ring.cljc
opqdonut, line 288: please also update `doc/ring/static.md`

opqdonut (changes requested):
Looks good, but I'd like some doc improvements/additions

modules/reitit-ring/src/reitit/ring.cljc
agorgl: Changed as requested

agorgl: Changed as requested

modules/reitit-ring/src/reitit/ring.cljc
agorgl: Changed as requested

agorgl:
@opqdonut This should be ready for merge

opqdonut (approved):

opqdonut:
Thanks a lot!

PR #727, opened 2025-03-26
Add option to disable index files served on paths that are not directories

agorgl:
Currently index files are served on non directory paths which is (by definition too) incorrect.
This PR adds a small guard to serve index files only for uris that are directories (end with a trailing slash).

### DIFF
--- a/doc/ring/static.md
+++ b/doc/ring/static.md
@@ -54,15 +54,16 @@ This way, they are only served if none of the actual routes have matched.
 
 `reitit.ring/create-file-handler` and `reitit.ring/create-resource-handler` take optionally an options map to configure how the files are being served.
 
-| key                | description |
-| -------------------|-------------|
-| :parameter         | optional name of the wildcard parameter, defaults to unnamed keyword `:`
-| :root              | optional resource root, defaults to `\"public\"`
-| :path              | path to mount the handler to. Required when mounted outside of a router, does not work inside a router.
-| :loader            | optional class loader to resolve the resources
-| :index-files       | optional vector of index-files to look in a resource directory, defaults to `[\"index.html\"]`
-| :index-redirect?   | optional boolean: if true (default), redirect to index file, if false serve it directly
-| :not-found-handler | optional handler function to use if the requested resource is missing (404 Not Found)
+| key                 | description |
+| --------------------|-------------|
+| :parameter          | optional name of the wildcard parameter, defaults to unnamed keyword `:`
+| :root               | optional resource root, defaults to `\"public\"`
+| :path               | path to mount the handler to. Required when mounted outside of a router, does not work inside a router.
+| :loader             | optional class loader to resolve the resources
+| :index-files        | optional vector of index-files to look in a resource directory, defaults to `[\"index.html\"]`
+| :index-redirect?    | optional boolean: if true (default), redirect to index file, if false serve it directly
+| :canonicalize-uris? | optional boolean: if true (default), try to serve index files for non directory paths (paths that end with slash)
+| :not-found-handler  | optional handler function to use if the requested resource is missing (404 Not Found)
 
 
 ### TODO
--- a/modules/reitit-ring/src/reitit/ring.cljc
+++ b/modules/reitit-ring/src/reitit/ring.cljc
@@ -227,11 +227,12 @@
    ;; TODO: ring.middleware.head/wrap-head
    ;; TODO: handle etags
    (defn -create-file-or-resource-handler
-     [response-fn {:keys [parameter root path loader allow-symlinks? index-files index-redirect? paths not-found-handler]
+     [response-fn {:keys [parameter root path loader allow-symlinks? index-files index-redirect? canonicalize-uris? paths not-found-handler]
                    :or {parameter (keyword "")
                         root "public"
                         index-files ["index.html"]
                         index-redirect? true
+                        canonicalize-uris? true
                         paths (constantly nil)}}]
      (let [options {:root root
                     :loader loader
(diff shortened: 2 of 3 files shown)


agorgl:
This PR is about modifying the behavior of file/resource handling (specifically in the context of index files), not routes in general. The slash handler does the opposite of what the current PR aims to achieve, it allows a route without a trailing slash to serve the same thing as a similar route that contains the trailing slash (so it does not apply to this problem).

The concept of an `index` file is commonly referred to as a `directory index`, and is a file that should be served when the url points to a directory (url has a trailing slash) as opposed to a file (url doesn't have a trailing slash). It is almost universal that when serving files/resources a url like `/doc/sample` should try to find a sample file in `/doc/` directory and a url like `/doc/sample/` should try to find an index file (whatever the configured index file patterns are) in the `/doc/sample/` directory and serve this.

With this change, reitit will now have a much more predictable behavior when serving files/resources with index files enabled for directories.

Deraen:
Some common examples:

Caddy will redirect from /foo to /foo/ if foo is a directory. It probably doesn't matter if there is an index file or not, at least if directory listing is enabled.
Nginx does the same (depends on location block and try_files etc. options).
Amazon CloudFront does also redirect /foo to /foo/ when foo is a directory.

Both do allow disabling this behavior, but it is the default (caddy disable_canonical_uris option.)

I think we should further document how Reitit works here, and maybe provide additional option to control this (especially if slash-handler is not enough for control this case also.) But I don't see completely disabling this as a reasonable change.

---

I'm not sure how much functionality we do want to include in this handler, users can just write their own handlers for complex cases. But we could e.g. add a new `redirect-canonical?` (default true) to match Caddy logic. `index-redirect?` can be kept around, but I don't think it is very useful option.

Cases:

- dir
    - IF `index-redirect?` redirects to `dir/index.html`
    - IF `redirect-canonical?` redirects to `dir/`
    - serves index.html
- dir/
    - IF `index-redirect?` redirects to `dir/index.html`
    - serves index.html
- dir/index.html
    - serves index.html
- dir/index.html/
    - maybe: IF `redirect-canonical?` redirects to `dir/index.html` (Caddy does this, CloudFront doesn't)
    - currently doesn't return anything?



https://caddyserver.com/docs/caddyfile/directives/file_server

agorgl:
The `index-redirect?` is mostly a cosmetic option to allow serving index files without redirecting to them and showing them in the url (that isn't very common practice nowdays).

I'll add the `redirect-canonical?` as it seems a good way to toggle the aforementioned functionality

agorgl:
Implemented the `redirect-canonical?` option as discussed, but I think the term redirect is not very good, as in the case `index-redirect?` option is false no redirect is performed. Maybe `canonicalize-uris?` is a better contestant?

Deraen:
I think we should also make `index-redirect?` default to false. It is kind of strange behavior to default to.

`canonicalize-uris?` could be a good name. I can change it when merging this.

Deraen:
The `redirect-canonical?` option doesn't yet work quite how I envisioned it.

I don't see the option where `dir` returns 404 if the `dir` exists and is a directory a commonly used behavior. Either it redirects to `dir/` (the canonical redirect option) or `dir/index.html` (index-redirect? option).

agorgl:
> I think we should also make `index-redirect?` default to false. It is kind of strange behavior to default to.
Yes I totally agree. I introduced it the previous PR where I tried to avoid breaking compatibility, I'll make another PR to change the default behavior.

> `canonicalize-uris?` could be a good name. I can change it when merging this.
I'll make the change now to make this more clear

agorgl:
Renamed `redirect-canonical?` to `canonicalize-uris?`.

agorgl:
> The `redirect-canonical?` option doesn't yet work quite how I envisioned it.
> 
> I don't see the option where `dir` returns 404 if the `dir` exists and is a directory a commonly used behavior. Either it redirects to `dir/` (the canonical redirect option) or `dir/index.html` (index-redirect? option).

When serving both (api?) routes and resources/files from the same base path, there is a case that an empty directory would redirect to a valid (api) route and this could be undesirable. I tried to avoid doing these kinds with the `canonicalize-uris?` option and let the functionality be only about file/resource index handling.

agorgl:
> > When serving both (api?) routes and resources/files from the same base path, there is a case that an empty directory would redirect to a valid (api) route and this could be undesirable.
> 
> In this case, user will just have the API routes first on the route tree, and those will return a response before file/resource handler is called.

I'm not sure I'm getting what do you want differently as a functionality from this.

The problematic case I'm describing is this:
- A regular route is served in `/something/`
- Resources/files are configured to be served from `/` (as a fallback if no regular route matches)
- Somebody tries the `/something` route
- Instead of getting a 404, it will get the `/something/` regular route
This conceptually is kinda wrong because a resource/file handler leads to a regular route and kinda crosses the boundaries of what a resource/handler should actually do.

If somebody needs to treat `/something` the same way as `/something/` there is the slash handler for that I suppose.

Deraen:
> Instead of getting a 404, it will get the /something/ regular route

This redirect only happen if there is `something/index.html` at the configured path in classpath/filesystem?

But I think I can now agree on returning index.html content only if `canonicalize-uris?` option is enabled when requesting `/dir/`.

I added just the redirect from `dir` to `dir/` when `dir/index.html` when `canonicalize-uris?` is enabled. Kind of similar to `index-redirect?` but the index filename isn't added to the path: https://github.com/metosin/reitit/pull/729/files

It would be nice to just leave this to slash-handler, but that handler isn't enabled usually, and a case that IS important we don't break is that if user types e.g. `example.com/api-docs` to access Swagger UI, we either show or redirect to the correct path. (We could adjust the options swagger-ui handler uses to create the resource handler, but right now it depends on redirect from this handler.)

agorgl:
I just saw your modifications, In summary this is what you propose:
```
diff --git a/modules/reitit-ring/src/reitit/ring.cljc b/modules/reitit-ring/src/reitit/ring.cljc
index 6fb9ce6a..15e953a8 100644
--- a/modules/reitit-ring/src/reitit/ring.cljc
+++ b/modules/reitit-ring/src/reitit/ring.cljc
@@ -259,8 +259,14 @@
                                           (loop [[file & files] index-files]
                                             (if file
                                               (if-let [resp (response (join-paths path file))]
-                                                (if index-redirect?
+                                                (cond
+                                                  index-redirect?
                                                   (response/redirect (join-paths uri file))
+
+                                                  (not (str/ends-with? uri "/"))
+                                                  (response/redirect (str uri "/"))
+
+                                                  :else
                                                   resp)
                                                 (recur files)))))))
            handler (if path
```

I think I agree, maybe we should change the `(not (str/ends-with? uri "/"))` condition to `(and canonicalize-uris? (not (str/ends-with? uri "/")))`?

Deraen:
The `canonicalize-uris?` check is still around `loop` so if uri doesn't end in `/`, the option is true. But yes, it could be easier to read if the check was also included in the `cond`.

agorgl:
Incorporated requested changes in this PR, will make another one to change the default value of `index-redirect?`.
This should be ready for merging

agorgl:
Commit for changing index-redirect? default value for false is ready here: https://github.com/agorgl/reitit/commit/73eaeac927e5b100e3f7f54e8b725d09b0e89073

Will make the PR after this one merges to master

PR #730, opened 2025-03-28
Change default index-redirect? value to false

agorgl:
Continuation of https://github.com/metosin/reitit/pull/727


Deraen:
Thanks! I can update the changelog and see if these changes should be reflected in the docs. I might even cut a new release today.

agorgl:
Fantastic! I'll upgrade my projects to the new release too when it comes out!

Issue #691, opened 2024-08-19
Unexpected handling of empty body response.


bombaywalla:
Hi folks

The following code does not behave as expected.

```
  (def test-handler
    (ring/ring-handler
     (ring/router [["/openapi.json"
                    {:get {:no-doc true
                           :openapi {:info {:title "testreitit"
                                            :description "Test Open API spec"
                                            :version "0.0.1"}}
                           :handler (openapi/create-openapi-handler)}}]
                   ["/test/{id}"
                    {:get {:description "id < 5 ==> 400 (null body); otherwise 200 and JSON"
                           :handler (fn [{{{:keys [id]} :path} :parameters}]
                                      (if (< id 5)
                                        {:status 400}
                                        {:status 200
                                         :body {:a 1}}))
                           :parameters {:path [:map [:id int?]]}
                           :responses
                           {:default {:description "default is json"
                                      :content {"application/json" {:schema [:map
                                                                             {:closed false}
                                                                             [:a int?]]}}}
                            400 {:description "400 and empty response"
                                 #_#_:content {:default {:schema nil?}}}}}}]]
                  {:validate spec/validate
                   :data {:coercion reitit.coercion.malli/coercion
                          :muuntaja m/instance
                          :middleware [openapi/openapi-feature
                                       swagger/swagger-feature
                                       parameters/parameters-middleware
                                       muuntaja/format-negotiate-middleware
                                       muuntaja/format-response-middleware
                                       exception/exception-middleware
                                       muuntaja/format-request-middleware
                                       coercion/coerce-response-middleware
                                       coercion/coerce-request-middleware
                                       multipart/multipart-middleware]}})
     (ring/routes (swagger-ui/create-swagger-ui-handler
                   {:path "/"
                    :config {:validatorUrl nil
                             :operationsSorter "alpha"
                             :urls [{:name "openapi" :url "/openapi.json"}]
                             :urls.primaryName "openapi"}} )
                  (ring/create-default-handler))))
  (def testx (jetty/run-jetty #'test-handler {:port 7777 :join? false}))
  (.stop ^org.eclipse.jetty.server.Server testx)
```

When I call the endpoint with a GET of `/test/2`, I get a 500 response code with the following body

```
{
  "value": null,
  "type": "reitit.coercion/response-coercion",
  "coercion": "malli",
  "in": [
    "response",
    "body"
  ],
  "humanized": [
    "invalid type"
  ]
}
```

I was expecting a 400 response code with an empty body.

[...]

PR #735, opened 2025-04-11
Resurrect :responses :default

opqdonut:


test/cljc/reitit/ring_coercion_test.cljc
```diff
@@ -680,6 +680,77 @@
...
+               {:b s/Int}]
+              [spec/coercion
+               {:a int?}
+               {:b int?}]]]
+       (testing (str coercion)
+         (let [app (ring/ring-handler
+                    (ring/router
+                     ["/foo" {:post {:responses {200 {:content {:default {:schema schema-200}}}
+                                                 201 {:content {"application/edn" {:schema schema-200}}}
+                                                 202 {:description "status code explicitly mentioned, but no :content map"}
```
opqdonut, line 700: The interaction here between `:content` and `:default` will probably mislead users.

- If a status code is listed, but has no `:content`, the coercion for the `:default` status will apply
- If a response code has a `:content`, it implicitly means non-mentioned content-types will have _no coercion at all_

It would be clearer, IMO, to have the `:default` response only apply when the status code isn't listed. That is, in this case, status 202 should have no response coercion at all, for any content type.

opqdonut:
This is currently on ok fix for the feature we broke in #715, but I think there's an opportunity to clarify even more stuff here.

We could also address #691

test/cljc/reitit/ring_coercion_test.cljc
bombaywalla: Yeah, I agree that the `:default` response should only apply if there is an unlisted status code.

As in #691 , it would be good to explicitly and separately specify and document the recommended way of specifying that an empty response body is expected.

lispyclouds:
Hey @opqdonut would there be a plan to release this? Hoping to get to update to the new reitit on some of my projects, not urgent but great to have it! Thanks again!

opqdonut:
I'll try to have some time to look at this, latest next week. Feel free to ping me if you don't hear from me.

opqdonut:
Jotting down some notes for myself.

Lookup order for coercions should be:

1. find status code in :responses, or if not present, look at :responses :default 
2. from there, find the content-type in :content, or look at :content :default, or :body
3. perform the coercion specified by :schema (or, in the case of :body, the value of :body). if there is no schema, do no coercion.

However, we should still be able to disable coercion with something like `(reitit.coercion.malli/create {:enabled false})`. This works currently by returning `nil` for the coercer, so `nil` means "don't coerce". However, other places use `-identity-coercer` to mean "don't coerce."

bombaywalla:
In Reitit, how do I specify that the :body must be empty?  (in contrast to ":body may be anything").

opqdonut:
Currently, the only way to do that is to use a schema that admits only `nil`. So for example `{:schema nil?}` works for malli & spec.

opqdonut:
I think I managed to cover all the corners. The default resolution order is now like I planned in my message above (see also: doc changes in the commits)

lispyclouds:
Looks great to me, thanks a lot for the effort!

mattiuusitalo (approved):

PR #746, opened 2025-06-18
Add :on-invalid for custom handling of invalid values in coercer

jgdavey:
In a transitional system (one going from no coercion at the edge to reitit + malli at the edge), it's useful to allow invalid requests and responses. In this way, we can, for example, allow invalid requests through (like with :validate option), but have a hook to log the validation errors. This "soft-coercion" method can help during the transition. Once there's more confidence that the data flowing in and out is valid, we can drop the custom handler and use the normal CoercionError path.

### DIFF
--- a/modules/reitit-malli/src/reitit/coercion/malli.cljc
+++ b/modules/reitit-malli/src/reitit/coercion/malli.cljc
@@ -78,8 +78,9 @@
               (if-let [coercer (get-coercer format)]
                 (if (-validate coercer transformed)
                   (-encode coercer transformed)
-                  (let [error (-explain coercer transformed)]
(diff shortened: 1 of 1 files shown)


opqdonut:
I talked this through with @Deraen and we don't feel like this is a big enough use case to justify the added complexity.

If you want to continue pursuing this, we would be ok with some sort of minimal extension point in reitit, that would allow you to write something like this in user code. Perhaps `->coercer` could be overridden? Then you could always return `true` from `-validate` and log the error.

If others are interested in this feature, please comment!

jgdavey:
Thank you for the response! Yes, I ended up extending the protocol and have something that works.

PR #779, opened 2026-02-09
Add test-doc-blocks

Deraen:
I suspected the malli coercion custom registry example was broken, so I wanted to get it tested and the coercion Malli options indeed had an extra layer of map.

TODO:
- [ ] Run on CI
- [ ] Enable for more doc files
- [x] I would like to use matcher-combinators or something to write the assertions, as there are many cases where return values include some records or reify values, or Clojure.spec instances which are different on each run.


Deraen:
Updated with forked test-doc-blocks. I didn't have time yet to go through every doc file.

PR #783, opened 2026-04-27
fix(reitit-malli): handle :and/:or composite schemas in OpenAPI :para…

MBS-Muthiah:
# fix(reitit-malli): handle `:and`/`:or` composite schemas in OpenAPI `:parameters`

## Problem

Composite Malli schemas like `[:and ...]` or `[:or ...]` used as route `:parameters` were **silently dropped** from the generated OpenAPI spec.

**Example — before fix:**

```clojure
:parameters {:query [:and [:map [:x int?]] [:map [:y string?]]]}
```

Generated OpenAPI parameters: `[]` ← empty, both `:x` and `:y` dropped

---

## Root Cause

In `modules/reitit-malli/src/reitit/coercion/malli.cljc`, the `-get-model-apidocs` method passed composite schemas directly to `json-schema/transform`, which returned:

- `[:and ...]` → `{:allOf [{...} {...}]}` — no top-level `:properties`
- `[:or ...]` → `{:anyOf [{...} {...}]}` — no top-level `:properties`

The caller in `modules/reitit-openapi/src/reitit/openapi.clj` destructured `{:keys [properties required]}` from the result, got `nil` for both, and produced an empty parameters list — **no error, no warning, just missing parameters**.

This did not affect Swagger because Swagger uses a different code path (`coercion/-get-apidocs`) that never does property-flattening itself.

---

## Fix

In `-get-model-apidocs`, before calling `json-schema/transform`, composite schemas are resolved into a single flat `:map`:

| Schema | Behaviour | Reason |
|---|---|---|
| `[:and [:map [:x int?]] [:map [:y string?]]]` | Merge all child `:map`s via `mu/merge`. All keys **required**. | All branches must match simultaneously |
| `[:or [:map [:x int?]] [:map [:y string?]]]` | Merge all child `:map`s via `mu/merge`, then wrap with `mu/optional-keys`. All keys **optional**. | Only one branch needs to match |
| `[:map ...]` | Unchanged — existing behaviour | — |

**After fix:**

```clojure
;; :and — both required
[{:in "query", :name :x, :required true,  :schema {:type "integer"}}
 {:in "query", :name :y, :required true,  :schema {:type "string"}}]

;; :or — both optional
[{:in "query", :name :x, :required false, :schema {:type "integer"}}
 {:in "query", :name :y, :required false, :schema {:type "string"}}]
```

---

## Files Changed

- `modules/reitit-malli/src/reitit/coercion/malli.cljc` — core fix in `-get-model-apidocs`
- `test/cljc/reitit/openapi_test.clj` — new `composite-malli-parameter-test` covering both `:and` and `:or` cases

---

## Tests

```
lein test reitit.openapi-test
→ Ran 14 tests containing 118 assertions. 0 failures, 0 errors.
```

### DIFF
--- a/modules/reitit-malli/src/reitit/coercion/malli.cljc
+++ b/modules/reitit-malli/src/reitit/coercion/malli.cljc
@@ -12,6 +12,28 @@
             [reitit.coercion :as coercion]))
 
 ;;
+;; helpers
+;;
+
+(defn- resolve-parameter-schema
+  "Resolves a Malli schema to a :map schema suitable for OpenAPI parameter generation.
+  :merge and :union are ref-like schemas — m/deref-all flattens them to :map automatically.
+  For :and/:or schemas, extracts the sole :map child when there is exactly one
+  (e.g. [:and [:map ...] [:fn ...]]). If there are multiple :map children the
+  original schema is returned unchanged so the caller's WARNING still fires."
+  [schema]
+  (let [resolved (m/deref-all schema)
+        schema-type (m/type resolved)]
+    (if (#{:and :or} schema-type)
+      (let [map-children (->> (m/children resolved)
+                              (map m/deref-all)
+                              (filter #(= :map (m/type %))))]
+        (if (= 1 (count map-children))
+          (first map-children)
+          resolved))
+      resolved)))
+
+;;
 ;; coercion
 ;;
 
@@ -146,8 +168,9 @@
            :openapi (if (= :parameter (:type options))
                       ;; For :parameters we need to output an object schema with actual :properties.
                       ;; The caller will iterate through the properties and add them individually to the openapi doc.
-                      ;; Thus, we deref to get the actual [:map ..] instead of some ref-schema.
-                      (let [should-be-map (m/deref model)]
+                      ;; resolve-parameter-schema dereferences ref-like schemas (:merge, :union) via m/deref-all
+                      ;; and also unwraps :and/:or to their sole :map child when present.
+                      (let [should-be-map (resolve-parameter-schema model)]
                         (when-not (= :map (m/type should-be-map))
                           (println "WARNING: Unsupported schema for OpenAPI (expected :map schema)" (select-keys options [:in :parameter]) should-be-map))
                         (json-schema/transform should-be-map (merge opts options)))
--- a/test/cljc/reitit/openapi_test.clj
+++ b/test/cljc/reitit/openapi_test.clj
@@ -3,6 +3,7 @@
             [clojure.test :refer [deftest is testing]]
             [jsonista.core :as j]
             [malli.core :as mc]
+            [malli.util :as mu]
             [matcher-combinators.test :refer [match?]]
             [muuntaja.core :as m]
             [reitit.coercion.malli :as malli]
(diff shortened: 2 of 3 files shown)


test/cljc/reitit/openapi_test.clj
```diff
@@ -1028,6 +1028,49 @@
...
+  (testing ":and schema in :parameters — both keys emitted, both required"
+    (let [app (ring/ring-handler
+               (ring/router
+                [["/openapi.json"
+                  {:get {:no-doc true
+                         :openapi {:info {:title "" :version "0.0.1"}}
+                         :handler (openapi/create-openapi-handler)}}]
+                 ["/and-query"
+                  {:get {:coercion malli/coercion
+                         :parameters {:query [:and
```
opqdonut: Since `-get-model-apidocs` runs `m/deref` on the schema, you can get the behaviour you want here with

```
{:query [:merge [:map [:x int?]] [:map [:y string?]]]}
```

This is because

```
user=> (require '[malli.core :as m])
user=> (require '[malli.registry :as mr])
user=> (require '[malli.util :as mu])
user=> (mr/set-default-registry! (mr/composite-registry (m/default-schemas) (mu/schemas)))
user=> (m/deref [:merge [:map [:x int?]] [:map [:y string?]]])
[:map [:x int?] [:y string?]]
```

test/cljc/reitit/openapi_test.clj
```diff
@@ -1028,6 +1028,49 @@
...
+  (testing ":or schema in :parameters — both keys emitted, both optional"
+    (let [app (ring/ring-handler
+               (ring/router
+                [["/openapi.json"
+                  {:get {:no-doc true
+                         :openapi {:info {:title "" :version "0.0.1"}}
+                         :handler (openapi/create-openapi-handler)}}]
+                 ["/or-query"
+                  {:get {:coercion malli/coercion
+                         :parameters {:query [:or
```
opqdonut: similarly to `:merge`, it should be possible to use `:union` here, but the keys might get accidentally marked as required

opqdonut (changes requested):
Thanks for the proposal! I think we should have a documented way of combining schemas for `:query` params. Please investigate how far you can get with `:merge` and `:union`, that might lead to a smaller diff needed.

test/cljc/reitit/openapi_test.clj
MBS-Muthiah: You're absolutely right—:merge/:union is cleaner. We initially went with :and/:or thinking it would avoid API changes, but if :merge/:union is preferred, we're happy to refactor.

One note: i've implemented the fix in openapi.clj rather than in the coercion layer (which was our first instinct), so the OpenAPI-specific handling is all in one place.

test/cljc/reitit/openapi_test.clj
MBS-Muthiah: You're absolutely right—:merge/:union is cleaner. We initially went with :and/:or thinking it would avoid API changes, but if :merge/:union is preferred, we're happy to refactor.

One note: i've implemented the fix in openapi.clj rather than in the coercion layer (which was our first instinct), so the OpenAPI-specific handling is all in one place.

opqdonut (changes requested):
I feel like you missed the meaning of my previous comment. What I meant was:

* without making any changes in the implementation
* write a test to see how much of the desired behaviour you can get by using `{:parameters {:query [:merge ...]}}` and `{:parameters {:query [:union ...]}}`

Also, I feel like I'm talking to an LLM. Since I am using my human attention to review this, please do me the courtesy of using your human attention to read & respond to my comments.

MBS-Muthiah:
> I feel like you missed the meaning of my previous comment. What I meant was:
> 
> * without making any changes in the implementation
> * write a test to see how much of the desired behaviour you can get by using `{:parameters {:query [:merge ...]}}` and `{:parameters {:query [:union ...]}}`
> 
> Also, I feel like I'm talking to an LLM. Since I am using my human attention to review this, please do me the courtesy of using your human attention to read & respond to my comments.

Hey,thanks for the feedback, this is my first opensource contribution, I am in my learning phase,I'll improve myself, 'll come back with a better test case exploring the {:parameters {:query [:merge ...]}} and {:parameters {:query [:union ...]}} approach.

MBS-Muthiah:
:merge works correctly. It flattens both branches into one map and the resulting :required list is the union of all required fields — which is precisely the right behaviour for :merge.

:union did not work correctly without a code change. Because the router pre-compiles schemas at startup, by the time the OpenAPI spec is generated the :union schema is no longer a raw [:union ...] vector — it's a compiled malli Schema object. The existing detection code used sequential? which only works on raw vectors, so it always fell through and treated :union identically to :merge (marking all fields required)
A one-line fix was needed in openapi.clj to use malli.core/type instead, which works correctly on compiled schemas.

The test is the proof — 28 assertions, 0 failures after the fix.

modules/reitit-openapi/src/reitit/openapi.clj
```diff
@@ -104,8 +104,34 @@
...
+                         ;;   :and   → {:allOf [...]} — required if required in any branch (union)
+                         ;;   :or    → {:anyOf [...]} — required only if required in ALL branches (intersection)
+                         ;;   :union → flat {:properties ...} from malli.util — branches must be processed
+                         ;;            individually to recover intersection semantics (a key is required
+                         ;;            only if it is required in every branch)
+                         ;; schema may be a raw vector [:union ...] or a compiled malli Schema object.
+                         ;; sequential? only works for raw vectors; use malli.core/type for compiled schemas.
+                         union-branches (let [schema-type (if (sequential? schema)
+                                                            (first schema)
+                                                            (try ((requiring-resolve 'malli.core/type) schema)
```
opqdonut: you can't use malli here, this code is meant to work for all coercion providers

modules/reitit-openapi/src/reitit/openapi.clj
```diff
@@ -104,8 +104,34 @@
...
+                                                                 (catch Exception _ nil)))]
+                                          (when (= :union schema-type)
+                                            (let [children (if (sequential? schema)
+                                                             (rest schema)
+                                                             ((requiring-resolve 'malli.core/children) schema))]
```
opqdonut: ditto

test/cljc/reitit/openapi_test.clj
```diff
@@ -1028,6 +1029,341 @@
...
+      ;; student-id only required in branch 1 → not in intersection → required: false
+      (is (match? {:required false} (by-name "student-id")))
+      ;; library-pass only required in branch 2 → not in intersection → required: false
+      (is (match? {:required false} (by-name "library-pass")))
+      ;; All 4 keys must appear
+      (is (= #{"hall-pass" "student-id" "teacher-name" "library-pass"}
+             (set (keys by-name))))
+      (is (nil? (validate spec)))))
+
+  (testing ":and schema with :fn — school-headers-and: required pass fields + optional identity + :fn validation"
```
opqdonut: why is this under the merge-union test?

test/cljc/reitit/openapi_test.clj
```diff
@@ -1028,6 +1029,341 @@
...
+      (is (match? {:required false} (by-name "library-pass")))
+      ;; Optional note fields
+      (is (match? {:required false} (by-name "principal-note")))
+      (is (match? {:required false} (by-name "classroom-note")))
+      ;; All 6 keys must appear
+      (is (= #{"hall-pass" "student-id" "library-pass" "teacher-name" "principal-note" "classroom-note"}
+             (set (keys by-name))))
+      (is (nil? (validate spec)))))
+
+  (testing ":or schema — school-headers-or: student-id path OR library-pass path, non-shared keys optional"
```
opqdonut: ditto

opqdonut (changes requested):
I see you've made some changes, but this is still not what I was looking for at this point.

I want to see what a minimal change that makes `:merge` and `:union` work would look like. The change shouldn't consider `:and` and `:or` at all at first. Then we can try adding something on top if the merge/union approach looks good to go.

Also, we can't have hacks like `requiring-resolve`, the fix needs to fit the current architecture with malli-specific stuff in `reitit.coercion.malli`.

Additionally, the set of test cases is good, but a bit redundant. Try to build a minimal set of test cases that showcases the merge/union logic you've built.

Thanks for your effort! There's still some distance to cover!

modules/reitit-malli/src/reitit/coercion/malli.cljc
```diff
@@ -11,6 +11,28 @@
...
+  (let [resolved (m/deref-all schema)
+        schema-type (m/type resolved)]
+    (if (#{:and :or} schema-type)
+      (let [map-children (->> (m/children resolved)
+                              (map m/deref-all)
+                              (filter #(= :map (m/type %))))]
+        (if (= 1 (count map-children))
+          (first map-children)
+          resolved))
+      resolved)))
```
opqdonut, line 34: I like this approach. It's a heuristic, but it's isolated to function, so it's easy to fix/extend later when we get user feedback.

opqdonut (approved):
Thanks! I think this is good to merge now. I'll merge it later today and add some docs & a changelog entry. Should be out in a release candidate really soon.

PR #785, opened 2026-05-19
Allow data: images in pedestal-swagger CSP

### LINKED ISSUE #447: In swagger examples: pedestal version the swagger icons are not showing because of CSP
Problem is found here https://github.com/metosin/reitit/blob/master/examples/pedestal-swagger/src/example/server.clj#L126

If you add this :img-src to the CSP settings it will start working
```
{:content-security-policy-settings
                                         {:default-src "'self'"
                                          :img-src "'self' data:" ; <-- allows the browser to load images from data: scheme
                                          :style-src "'self' 'unsafe-inline'"
                                          :script-src "'self' 'unsafe-inline'"}}
```

SAY-5:
Closes #447. Adds `:img-src "'self' data:"` to the pedestal-swagger example's secure-headers so Swagger UI's inline SVG checkbox icons render correctly. Mirrors the snippet @opqdonut said "PR welcome!" for in the issue thread.

### DIFF
--- a/examples/pedestal-swagger/src/example/server.clj
+++ b/examples/pedestal-swagger/src/example/server.clj
@@ -122,9 +122,10 @@
        ::server/join? false
        ;; no pedestal routes
        ::server/routes []
-       ;; allow serving the swagger-ui styles & scripts from self
+       ;; allow serving the swagger-ui styles, scripts and inline images from self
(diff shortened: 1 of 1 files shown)


opqdonut (approved):
thanks!

PR #791, opened 2026-08-19
fix(malli): handle :and, :or, and :any schemas as OpenAPI query params

MBS-Muthiah:
## Problem

When using Malli composite schemas as `:query` parameters in reitit routes, the OpenAPI spec
generator silently discarded the entire parameter block and printed
`WARNING: Unsupported schema for OpenAPI (expected :map schema)` to stdout.

Three patterns were broken:

### 1. `:and` with multiple `:map` schemas + `:fn` predicates

```clojure
:parameters {:query [:and
                     [:and
                      [:map [:from :string] [:to :string] [:status :string]]
                      [:map [:page {:optional true} :int] [:size {:optional true} :int]]]
                     [:fn (fn [{:keys [from to]}] (< from to))]]}
```

**Result:** all query parameters silently dropped from the generated spec.

### 2. `:or` with alternative `:map` schemas

```clojure
:parameters {:query [:or
                     [:map [:category :string]]
                     [:map [:item-id :uuid]]]}
```

**Result:** both parameters dropped. Routes where one of two keys is required had no spec output.

### 3. `:any` as a query schema

```clojure
:parameters {:query :any}
```

**Result:** `WARNING` printed, no parameters emitted, even though `:any` is a valid pass-through intent.

---

## Root Cause

`(-get-model-apidocs)` in `malli.cljc` expected the schema to resolve to a plain `:map` after
`m/deref`. Composite schemas (`:and`, `:or`, `:any`) were not `:map`, so `json-schema/transform`
produced no `:properties` and the parameter loop emitted nothing.

---

## Solution

Added two private helpers in `malli.cljc`:

- **`collect-and-maps`** — recursively walks an `:and` tree collecting all `:map` children,
  ignoring `:fn` predicates (cross-field constraints with no OpenAPI equivalent)
- **`resolve-parameter-schema`** — normalises a schema to a single `:map` before transformation:
  - `:and` → recursively collects all `:map` descendants, merges them with `mu/merge`
  - `:or` → unions all direct `:map` children via `mu/merge`, marks every key optional with
    `mu/optional-keys` (only one branch need satisfy)
  - `:any` → returns empty `[:map]` (open pass-through, no specific parameters emitted, no warning)
  - `:merge` / `:union` → already handled upstream by `m/deref-all`

---

## Files Changed

| File | Change |
|------|--------|
| `modules/reitit-malli/src/reitit/coercion/malli.cljc` | Add `collect-and-maps` + `resolve-parameter-schema`; wire into `(-get-model-apidocs)` |
| `test/cljc/reitit/openapi_test.clj` | Add `malli-composite-query-params-test` covering flat `:and`, nested `:and`, `:or`, and `:any` |

### DIFF
--- a/doc/ring/openapi.md
+++ b/doc/ring/openapi.md
@@ -228,9 +228,10 @@ supports the following forms as well:
 ```clj
 [:union [:map ...] [:map ...]]
 [:merge [:map ...] [:map ...]]
-[:and [:map ...] [:fn ...]]
-[:and [:map ...] :anything-that's-not-a-map]
-[:or [:map ...] :anything-that's-not-a-map]
+[:and [:map ...] [:map ...] :anything-that's-not-a-map]
+[:and [:map ...] [:and [:map ...] [:map ...]]]
+[:or [:map ...] [:map ...] :anything-that's-not-a-map]
+[:or [:map ...] [:or [:map ...] [:map ...]]]
 ```
 
 This support is only for Malli so far, not Plumatic Schema or Spec.
--- a/modules/reitit-malli/src/reitit/coercion/malli.cljc
+++ b/modules/reitit-malli/src/reitit/coercion/malli.cljc
@@ -15,22 +15,30 @@
 ;; helpers
 ;;
 
+(defn- collect-and-maps
+  "Recursively collect :map schemas from an :and tree, ignoring :fn and other non-map entries."
+  [schema]
+  (let [s (m/deref-all schema)]
+    (case (m/type s)
+      :map [s]
+      :and (mapcat collect-and-maps (m/children s))
+      [])))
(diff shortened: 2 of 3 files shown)


opqdonut:
Did you forget to commit malli-composite-query-params-test ? I don't see any new test cases here. I like the idea, but I'll need to think about it for a while.

MBS-Muthiah:
> Did you forget to commit malli-composite-query-params-test ? I don't see any new test cases here. I like the idea, but I'll need to think about it for a while.

yeahh i missed it, will try to add new test cases for it

opqdonut (commented):
Thanks, this looks good. I'll reread on monday and probably merge then!

PR #794, opened 2026-09-04
Extend Expand to APersistentMap on JVM

The-Alchemist:
## Summary
- On JVM, extend `reitit.core/Expand` to `clojure.lang.APersistentMap` instead of the concrete `PersistentArrayMap` / `PersistentHashMap` classes.

Behavior for stock Clojure maps is unchanged (`expand` still returns `this`). Protocol dispatch continues to cache per concrete class after first resolve.

This keeps route-data map expansion working for any `APersistentMap` subclass (e.g. `PersistentTreeMap`, and alternate JVM Clojure runtimes that provide their own map implementations under that hierarchy), without requiring library-specific reader features.

## Reason
[Cloffle](https://github.com/The-Alchemist/cloffle-clojure) is a GraalVM/Truffle Clojure runtime with specialized map types that extend `APersistentMap` but not `PersistentArrayMap` / `PersistentHashMap`. Extending those concrete classes means Cloffle map literals miss `Expand`; targeting `APersistentMap` fixes that without changing stock Clojure behavior.

We  use `reitit`  as part of our compatibility tests with Clojure, and we maintain this patch for `reitit` internally.  This PR would help us avoid that, without any performance impact on reitit itself.

### DIFF
--- a/modules/reitit-core/src/reitit/core.cljc
+++ b/modules/reitit-core/src/reitit/core.cljc
@@ -16,12 +16,11 @@
      :cljs cljs.core.Keyword)
   (expand [this _] {:name this})
 
-  #?(:clj  clojure.lang.PersistentArrayMap
+  #?(:clj  clojure.lang.APersistentMap
      :cljs cljs.core.PersistentArrayMap)
   (expand [this _] this)
 
-  #?(:clj  clojure.lang.PersistentHashMap
-     :cljs cljs.core.PersistentHashMap)
+  #?(:cljs cljs.core.PersistentHashMap)
   (expand [this _] this)
 
   #?(:clj  clojure.lang.Fn


modules/reitit-core/src/reitit/core.cljc
```diff
@@ -16,12 +16,11 @@
...
   (expand [this _] {:name this})
 
-  #?(:clj  clojure.lang.PersistentArrayMap
+  #?(:clj  clojure.lang.APersistentMap
      :cljs cljs.core.PersistentArrayMap)
   (expand [this _] this)
 
-  #?(:clj  clojure.lang.PersistentHashMap
-     :cljs cljs.core.PersistentHashMap)
+  #?(:cljs cljs.core.PersistentHashMap)
```
opqdonut, line 23: hmm: how is this syntactically valid in the `clj` case? shouldn't the next line be inside the #? guard as well?

opqdonut (changes requested):

modules/reitit-core/src/reitit/core.cljc
The-Alchemist: i won't pretend to understand the details of the `#?(:clj` and `#?(:cljs)` conditionals, but i made a few tweaks:  https://github.com/metosin/reitit/pull/794/changes/fb92a9fc9983b7efb7da3161fef38e2ca7ad759d

should be easier to read, and added a test case for good measure.

let me know what you think!!

opqdonut (approved):
golden, thanks

PR #796, opened 2026-09-04
Extend reitit.walk keywordize to IPersistentVector

The-Alchemist:
## Summary
- `reitit.walk` extends `IKeywordize` to `PersistentVector` by exact class. `subvec` (`APersistentVector$SubVector`) and other `IPersistentVector` implementations therefore fall through to `Object` and are not walked, so `keywordize-keys` disagrees with `clojure.walk`.
- Extend `clojure.lang.IPersistentVector` instead, matching the existing `IPersistentMap` fallback. `PersistentVector` behavior is unchanged.
- Add a `keywordize-subvec` regression test (`gen/any-equatable` never produces subvecs).

## Motivation
[Cloffle](https://github.com/The-Alchemist/cloffle-clojure) is a Truffle/GraalVM implementation of Clojure. Its small vectors are `PersistentTuple` (`IPersistentVector`, not `PersistentVector`), so `reitit.walk/keywordize-keys` skipped nested maps inside `(vector …)` while `clojure.walk` did not. The same exact-class gap shows up on JVM Clojure with `subvec`. Extending `IPersistentVector` covers both.

### DIFF
--- a/modules/reitit-core/src/reitit/walk.clj
+++ b/modules/reitit-core/src/reitit/walk.clj
@@ -19,8 +19,10 @@
   [coll]
   (into (empty coll) keywordize-xf coll))
 
+;; Protocol extend is exact-class. IPersistentVector covers PersistentVector,
+;; subvec (APersistentVector$SubVector), and other IPersistentVector impls.
 (doseq [type [clojure.lang.PersistentHashSet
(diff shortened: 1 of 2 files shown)


opqdonut (approved):
reasonable, thanks

PR #797, opened 2026-09-04
doc: explain middleware execution order more explicitly

### LINKED ISSUE #790: Document middleware execution order
It would be nice if the order of execution of middleware in each `:middleware [,,, vector ,,,]` would be documented.

Tests show it by example, "by default, all middleware are applied in order" (even though I'd use a different wording than "in order", because it appears ambiguous):

https://github.com/metosin/reitit/blob/c1b9a105d3cea34ced07fa999b1ac07f343065b1/test/cljc/reitit/middleware_test.cljc#L238-L266

A sentence explaining it in the documentation in easy English could be:

> Chains of middleware specified in a vector are executed in sequential order, first to last.

A good place for adding a sentence or example might be:

https://github.com/metosin/reitit/blob/c1b9a105d3cea34ced07fa999b1ac07f343065b1/doc/ring/ring.md?plain=1#L221-L240

The example could be extended to:
```clj
(def app
  (ring/ring-handler
    (ring/router
      ["/api" {:middleware [[wrap :3-parent]]}
       ["/get" {:get handler
                :middleware [[wrap :4-route]]}]]
      {:data {:middleware [[wrap :2-top-level-route-data]]}})
    nil
    {:middleware [[wrap :1-top-a]
                  [wrap :1-top-b]]}))

(app {:request-method :get, :uri "/api/get"})
; {:status 200, :body [:1-top-a :1-top-b :2-top-level-route-data :3-parent :4-route :handler]}
```

Below the example I'd add:

> The same order within each vector is also applied in all other locations where middleware can be specified.

Another location where one might expect such information is the documentation of how data-driven middleware chains are created:

https://github.com/metosin/reitit/blob/c1b9a105d3cea34ced07fa999b1ac07f343065b1/doc/ring/data_driven_middleware.md?plain=1#L64-L87

It's already there in the example, but it's a bit hard to find by search, because it does not mention the word "order".  Also, it just says ambiguously that middleware is "applied correctly", but (the example aside) does not explain in words what order the "correct" order would be.

devurandom:
Closes: https://github.com/metosin/reitit/issues/790

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ pom.xml.asc
 /_book
 figwheel_server.log
 /.idea
+*.iml
 .clj-kondo
 .shadow-cljs
 .cache
--- a/doc/ring/data_driven_middleware.md
+++ b/doc/ring/data_driven_middleware.md
@@ -79,12 +79,15 @@ The following examples produce identical middleware runtime functions.
                        :handler handler}}]])))
 ```
 
-All the middlewares are applied correctly:
+All the middlewares are applied in sequential order, first to last,
+outside to inside:
 
 ```clj
 (app {:request-method :get, :uri "/api/ping"})
 ; {:status 200, :body [1 2 3 :handler]}
 ```
+
+See also the [execution order of ring middleware](ring.md#execution-order).
 
 ## Compiling middleware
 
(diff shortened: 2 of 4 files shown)


opqdonut (approved):

opqdonut:
Thanks!

PR #798, opened 2026-09-05
Make OpenAPI/Swagger tests independent of parameter array order

The-Alchemist:
## Summary
- OpenAPI and Swagger `:parameters` are unordered arrays. Tests previously used `=` / `match?` on those vectors, which is JVM-dependent because Clojure maps do not guarantee seq order when the spec is built from `:parameters` maps.
- Sort `:parameters` in the *tests* by `:in` and name before comparing. Production emission is unchanged.

## Motivation
Clojure's map contract does not promise insertion order except for `array-map`. Alternate JVM implementations (like [our Cloffle, based on Clojure on GraalVM](https://github.com/The-Alchemist/cloffle-clojure)) can emit `path` before `query`. Spec consumers and golden tests compare parameter *vectors*, which are order-sensitive.

We currently maintain a separate patch to the tests because we run `reitit` as part of our compatibility suite.

## Compatibility
No public API change and no change to generated OpenAPI/Swagger output. Only test helpers and assertions were updated.

### DIFF
--- a/modules/reitit-openapi/src/reitit/openapi.clj
+++ b/modules/reitit-openapi/src/reitit/openapi.clj
@@ -212,7 +212,7 @@
            strip-top-level-keys #(dissoc % :id :info :host :basePath :definitions :securityDefinitions)
            strip-endpoint-keys #(dissoc % :id :parameters :responses :summary :description)
            openapi (->> (strip-endpoint-keys openapi)
-                        (merge {:openapi "3.1.0"
+                        (merge {:openapi "3.2.0"
                                 :x-id ids}))
            accept-route (fn [route]
                           (-> route second :openapi :id (or ::default) (trie/into-set) (set/intersection ids) seq))
--- a/modules/reitit-openapi/src/reitit/openapi.clj
+++ b/modules/reitit-openapi/src/reitit/openapi.clj
@@ -78,6 +78,17 @@
 
 (def ^:private form-content-type "application/x-www-form-urlencoded")
 
+;; OpenAPI parameter locations in spec order. Clojure maps do not guarantee
+;; seq order, so emit :query/:header/:cookie/:path independently of the
+;; :parameters map implementation.
+(def ^:private parameter-in-order [:query :header :cookie :path])
+
+(defn- ordered-parameter-entries [parameters]
+  (keep (fn [in]
+          (when-let [schema (get parameters in)]
+            [in schema]))
+        parameter-in-order))
+
 (defn -get-apidocs-openapi
   [coercion {:keys [request muuntaja parameters responses openapi/request-content-types openapi/response-content-types]} definitions]
   (let [{:keys [body form multipart]} parameters
@@ -103,7 +114,7 @@
     (merge
      (when (seq parameters)
        {:parameters
-        (->> (for [[in schema] parameters
+        (->> (for [[in schema] (ordered-parameter-entries parameters)
                    :let [{:keys [properties required]} (->schema-object schema {:in in :type :parameter})
                          required? (partial contains? (set required))]
                    [k schema] properties]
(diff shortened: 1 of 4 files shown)


test/cljc/reitit/openapi_test.clj
```diff
@@ -480,6 +480,28 @@
         (testing "spec is valid"
           (is (nil? (validate spec))))))))
 
+(deftest parameter-location-order-independent-of-map-seq
```
The-Alchemist, line 483: overkill, maybe?

test/cljc/reitit/swagger_test.clj
```diff
@@ -444,6 +444,26 @@
     (is (= ["query" "body" "formData" "header" "path"]
            (map :in (get-in spec [:paths "/parameters" :post :parameters]))))))
 
+(deftest parameter-location-order-independent-of-map-seq
```
The-Alchemist, line 447: overkill?

modules/reitit-openapi/src/reitit/openapi.clj
```diff
@@ -201,7 +212,7 @@
            strip-top-level-keys #(dissoc % :id :info :host :basePath :definitions :securityDefinitions)
            strip-endpoint-keys #(dissoc % :id :parameters :responses :summary :description)
            openapi (->> (strip-endpoint-keys openapi)
-                        (merge {:openapi "3.2.0"
+                        (merge {:openapi "3.1.0"
```
Deraen: Merge/rebase artifact likely, openapi upgrade shouldn't be reverted.

The-Alchemist: good catch!  fixed!

opqdonut:
Quick comment: I'm not sure I like this hack. I feel that the ordering should get fixed at json-serialisation-time, not in the clojure data structures.

The-Alchemist:
@opqdonut 
> Quick comment: I'm not sure I like this hack. I feel that the ordering should get fixed at json-serialisation-time, not in the clojure data structures.

I think you're totally right; I fixed the tests to be sort-order independent.  let me know what you think!

test/cljc/reitit/openapi_test.clj
```diff
@@ -435,29 +461,31 @@
...
-                        :name "c"
-                        :required true
-                        :description "description :c"
-                        :schema {:type "string"}}
-                       {:in "path"
-                        :name "p"
-                        :required true
-                        :description "description :p"
-                        :schema {:type "string"}}]
+          (is (match? (sorted-parameters
```
opqdonut: bw matcher-combinators has in-any-order, which could take care of this

test/cljc/reitit/swagger_test.clj
```diff
@@ -495,18 +523,20 @@
...
+                  :name "file"
+                  :required true
+                  :type "file"}
+                 {:description ""
+                  :in "formData"
+                  :name "more"
+                  :required true
+                  :type "string"}])
+               (sorted-parameters
+                (normalize
```
opqdonut: sorted-parameters could be folded into normalize in this test namespace

test/cljc/reitit/swagger_test.clj
```diff
@@ -440,9 +467,10 @@
                ["/swagger.json"
                 {:get {:no-doc true
                        :handler (swagger/create-swagger-handler)}}]]))
-        spec (:body (app {:request-method :get, :uri "/swagger.json"}))]
-    (is (= ["query" "body" "formData" "header" "path"]
-           (map :in (get-in spec [:paths "/parameters" :post :parameters]))))))
+        spec (:body (app {:request-method :get, :uri "/swagger.json"}))
+        ins (map :in (get-in spec [:paths "/parameters" :post :parameters]))]
+    (is (= 5 (count ins)))
+    (is (= #{"query" "body" "formData" "header" "path"} (set ins)))))
```
opqdonut: btw, you could get by with one `is` by using `(is (= [...] (sort ...)))`

test/cljc/reitit/swagger_test.clj
```diff
@@ -23,6 +23,32 @@
...
+                        (reduce-kv
+                         (fn [ms method op]
+                           (assoc ms method
+                                  (cond-> op
+                                    (contains? op :parameters)
+                                    (update :parameters sorted-parameters))))
+                         methods
+                         methods)))
+               paths
+               paths)))))
```
opqdonut, line 51: I'm tempted to code-golf this, but I can do it after merging this PR 😁

opqdonut (commented):
The approach is good. I have some improvement suggestions on small details, but I can also implement those myself if you don't feel like making yet another revision of this. Let me know which way you want to do it!

opqdonut (approved):
thanks for the fixes!

PR #802, opened 2026-09-18
refactor openapi & swagger tests a bit

opqdonut:
follow-up to #798


opqdonut:
FYI @The-Alchemist

The-Alchemist:
looks much better :)

